PR #9, opened 2021-01-23
Update README.md

peterfisher:
I get `Unable to find a specification` without doing a `pod update` first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ npm install
 npm run setup
 gem install cocoapods
 cd ios
+pod update
 pod install
 ```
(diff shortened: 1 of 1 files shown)


peterfisher:
Actually the entire Pods directory (./Sefaria-iOS/ios/Pods) is causing a mess. I wasn't able to build at all getting all sorts of missing header file compilation issues. Once I `rm -rf ./Sefaria-iOS/ios/Pods` directory I was able to run `pod install` and get everything to compile. Maybe the Pods directory should be added to the .gitignore and it should be installed on each users machine via pod install.

PR #105, opened 2024-01-21
Alt versions

nsantacruz:
This PR does a few things

1. Upgrades us from React Native 71 to 73
2. Allows support for multiple offline versions of books, not just the default version


DownloadControl.js
```diff
@@ -557,7 +557,7 @@ async function requestNewBundle(bookList, badResponseWaitTime=3000) {
    * This method will continually ping the server until the download is ready (and a 200 is received).
    */
   while (true) {
-    let response = await fetch(`${DOWNLOAD_SERVER}/makeBundle`, {
+    let response = await fetch(`${DOWNLOAD_SERVER}/makeBundle?schema_version=${SCHEMA_VERSION}`, {
```
nsantacruz, line 560: old versions of the app that don't have this change will default to requesting v7 which is the version on the server and it will break their apps!

sefaria.js
```diff
@@ -220,7 +259,130 @@ Sefaria = {
...
+    try {
+      metadata = await Sefaria.loadOfflineSectionMetadata(ref);
+    } catch(error) {
+      throw ERRORS.OFFLINE_LIBRARY_NOT_COMPATIBLE_WITH_V7;
+    }
+    Sefaria._jsonSectionData[key] = metadata;
+    return metadata;
+  },
+  loadOfflineSectionMetadata: async function(ref) {
+    if (Sefaria.shouldLoadFromApi()) {
```
edamboritz, line 357: Is the awareness of online features such as the api perhaps to high level for a lower level functions like this?

edamboritz: Also, I think the dealing with file system level concerns such as unzipping should be done separately and this file left only to handle an better defined and existing metadata file per section.

edamboritz (commented):
This is truly amazing work to get an entirely new feature available! 

Some general comments:
 
I think we should make sure that all new functions have docstrings exaplining what they do 
I think we also have a problem here where there are too many similar function names. Its very confusing to understand waht each function does. We should consider modularization of "different area of concern" and groups of functions into separate files.

offline.js
```diff
@@ -0,0 +1,417 @@
...
+        return processed;
+    }
+    return {result: textFromRefData(processed)};
+};
+
+export const loadTextTocOffline = function(title) {
+    return _loadJSON(_JSONSourcePath(title + "_index"));
+}
+
+export const getOfflineVersionObjectsAvailable = function(ref) {
```
nsantacruz, line 29: move to Sefaria

edamboritz (approved):

PR #106, opened 2024-02-05
Connections panel

YishaiGlasner:
This PR is for redesign of the connections panel, mirroring the web layout.
It introduces significant changes to ConnectionPanel.jsx, introducing new components such as ConnectionsPanelSection, TopButtons, ResourcesList, and ToolsList. These components streamline the management of arrays of ToolsButton instances, whether with or without titles.


ConnectionsPanel.js
```diff
@@ -248,6 +257,7 @@ class ConnectionsPanel extends React.PureComponent {
                 withArrow={false}
                 buttonStyle={{margin: 0, padding: 0}}
                 key={cat.category}
+                isMainMenu={this.props.connectionsMode === null}
```
nsantacruz, line 260: I like the name of this prop. May as well define it before line 242 so you can use it to define `enText`.

ConnectionsPanel.js
```diff
@@ -272,25 +282,52 @@ class ConnectionsPanel extends React.PureComponent {
             }
           }
           if (this.props.connectionsMode === null) {
```
nsantacruz, line 284: use `isMainMenu` variable here as well.

ConnectionsPanel.js
```diff
@@ -272,25 +282,52 @@ class ConnectionsPanel extends React.PureComponent {
...
-                segmentRef={this.props.segmentRef}
-                heSegmentRef={this.props.heSegmentRef}
-                categories={this.props.categories}
-                shareCurrentSegment={this.props.shareCurrentSegment}
-                reportError={this.props.reportError}
-                viewOnSite={this.props.viewOnSite}
-                reloadRelated={this.reloadRelated}
-              />
-            );
+              const collapsedTopLevelLimit = 3;
```
nsantacruz, line 285: (1) the `default` case of the `switch` should be it's own component. There's a lot of logic here.
(2) the new component should rely on a hook `useLinkSummaryList` (not 100% on the name) which stores the state of `showAllRelated` and returns (a) a callback to toggle `showAllRelated`, (b) just the data that needs to be displayed (not the React components). This will separate data from view logic which is currently very mixed and confusing. I realize it was like this prior to your work but your work is building on it.

ConnectionsPanel.js
```diff
@@ -272,25 +282,52 @@ class ConnectionsPanel extends React.PureComponent {
...
+                  viewList.push((
+                      <ToolsButton
+                          onPress={this.onPresshShowMoreLess}
+                          text={strings.less}
+                          icon={iconData.get('up', this.props.themeStr)}
+                      />
+                  ));
+                }
+              }
+              viewList = [
```
nsantacruz, line 307: Don't reuse the same variable name

ConnectionsPanel.js
```diff
@@ -120,7 +128,7 @@ class ConnectionsPanel extends React.PureComponent {
       </View>
     );
     switch (this.props.connectionsMode) {
-      case 'filter': // fall-through
+      case 'filter': return; // fall-through
```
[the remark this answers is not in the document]
nsantacruz: What is this case? Why does ConnectionsPanel even render if `connectionsMode` is `filter` if it just returns?

sefaria.js
```diff
@@ -815,7 +815,7 @@ Sefaria = {
       return new Promise(function(resolve, reject) {
         // Returns an ordered array summarizing the link counts by category and text
         // Takes an array of links which are of the form { "category", "sourceHeRef", "sourceRef", "textTitle"}
-        var summary = {"All": {count: 0, books: {}, hasEn: false}, "Commentary": {count: 0, books: {}, hasEn: false}};
+        var summary = {"Commentary": {count: 0, books: {}, hasEn: false}};
```
nsantacruz, line 818: Since you're touching this anyway, change to `let`.

Styles.js
```diff
@@ -1239,13 +1239,13 @@ export default StyleSheet.create({
...
-    borderTopWidth: 1,
+    borderTopWidth: null,
     justifyContent: 'flex-start',
+    paddingHorizontal: 0,
   },
   toolsButtonIcon: {
     width: iPad ? readerSideMarginIpad : readerSideMargin,
-    alignItems: 'center',
-  },
+},
```
nsantacruz, line 1248: indent

Styles.js
```diff
@@ -1545,6 +1545,14 @@ export default StyleSheet.create({
     fontWeight: 'bold',
   },
   flex1: {
-    flex: 1,
-  },
+  flex: 1,
+  },
+connectionPanelSection: {
```
nsantacruz, line 1550: fix indent of next lines

Styles.js
```diff
@@ -1545,6 +1545,14 @@ export default StyleSheet.create({
...
+  paddingHorizontal: 40,
+},
+connectionPanelTitle: {
+  paddingVertical: 10,
+  borderBottomWidth: 1,
+  borderBottomColor: '#CCCCCC',
```
nsantacruz, line 1556: Remove this color from `Styles.js`. Should use the appropriate color in `themes`. In this case I see there is no matching theme style here. Create a new style in both `ThemeWhite.js` and `ThemeBlack.js`:

```js
lightGrayBorderBottom: {
  borderBottomColor: Colors.lightGrey,
}
```

Misc.js
```diff
@@ -593,23 +594,32 @@ const LibraryNavButton = ({
   hasEn,
   withArrow,
   buttonStyle,
+  isMainMenu
```
nsantacruz, line 597: the `isMainMenu` case feels different enough to warrant it's own component. There are 4 if statements checking this value.

ConnectionsPanel.js
```diff
@@ -319,92 +355,18 @@ class ConnectionsPanel extends React.PureComponent {
...
   const flexDir = interfaceLanguage === "english" ? null : styles.rtlRow;
   const hasIcon = !!icon;
   icon = icon || iconData.get('sheet', themeStr);  // default to arbitrary icon that will be invisible if icon wasn't passed
-  const iconComp = (<View style={[styles.toolsButtonIcon, hasIcon ? null : styles.readerNavSectionMoreInvisible]}><Image source={icon} style={styles.menuButton} resizeMode={'contain'}></Image></View>);
+  const iconComp = (<View style={[styles.toolsButtonIcon, flexDir, hasIcon ? null : styles.readerNavSectionMoreInvisible]}><Image source={icon} style={styles.menuButton} resizeMode={'contain'}></Image></View>);
   const countComp = !!count || count === 0 ? <Text style={[styles.enInt, theme.secondaryText, styles.spacedText]}>{` (${count}) `}</Text> : null
   return (
     <SefariaPressable
-      extraStyles={[styles.searchFilterCat, styles.toolsButton, flexDir, theme.bordered]}
+      extraStyles={[styles.searchFilterCat, styles.toolsButton, flexDir, theme.bordered, {height: 36}]}
```
nsantacruz, line 369: create a new style in Styles.js for the height. By passing a literal object as a style it forces the component to rerender whenever the parent rerenders.

ConnectionsPanel.js
```diff
@@ -420,4 +382,110 @@ ToolsButton.propTypes = {
   count:         PropTypes.number,
 };
 
+const ConnectionsPanelSection = ({ title, children }) => {
+  const { themeStr, interfaceLanguage } = useContext(GlobalStateContext);
+  const theme = getTheme(themeStr);
```
nsantacruz, line 387: `theme` exists in GlobalStateContext. `themeStr` is unnecessary in this case. See other usages of `useContext(GlobalStateContext);`

ConnectionsPanel.js
```diff
@@ -420,4 +382,110 @@ ToolsButton.propTypes = {
...
+  return (
+    <View style={styles.connectionPanelSection}  flexdir={interfaceLanguage === "english" ? null : styles.rtlRow}>
```
nsantacruz, line 389: I'm pretty sure `flexdir` isn't a valid prop on `View`. Is this logic working? Use `<FlexFrame/>` in Misc.js instead.

ConnectionsPanel.js
```diff
@@ -420,4 +382,110 @@ ToolsButton.propTypes = {
...
+      {title &&
+        <View style={styles.connectionPanelTitle}>
+          <Text style={[interfaceLanguage === "hebrew" ? styles.heInt : styles.enInt, theme.tertiaryText]}>
+            {title}
+          </Text>
+        </View>
+      }
+      <View>
```
nsantacruz, line 397: why is there an extra `<View>` here?

ConnectionsPanel.js
```diff
@@ -420,4 +382,110 @@ ToolsButton.propTypes = {
...
+  );
+};
+ConnectionsPanelSection.propTypes = {
+  title: PropTypes.string,
+  children: PropTypes.node.isRequired,
+};
+
+const TopButtons = ({relatedHasError, reloadRelated, themeStr, setConnectionsMode, sheet, versionsCount}) => {
+  return (
+    <ConnectionsPanelSection>
```
nsantacruz, line 410: Nice use of children!

ConnectionsPanel.js
```diff
@@ -420,4 +382,110 @@ ToolsButton.propTypes = {
...
+      {!relatedHasError ? null : <ToolsButton
```
nsantacruz, line 411: switch the order of this if statement so it is positive (not using `!`). This tends to be simpler to understand.

ConnectionsPanel.js
```diff
@@ -420,4 +382,110 @@ ToolsButton.propTypes = {
...
+    </ConnectionsPanelSection>
+  )
+};
+TopButtons.propTypes = {
+    themeStr:           PropTypes.string.isRequired,
+    setConnectionsMode: PropTypes.func.isRequired,
+    versionsCount:      PropTypes.number.isRequired,
+    relatedHasError:    PropTypes.bool,
+    reloadRelated:      PropTypes.func.isRequired,
+  // what is sheet?
```
nsantacruz, line 435: this is a good question. It's a bad habit from a long time ago. Sheet is the full sheet object, assuming you're viewing a sheet. We often use it as a boolean to say "are you currently viewing a sheet?". This is very bad practice...

ConnectionsPanel.js
```diff
@@ -420,4 +382,110 @@ ToolsButton.propTypes = {
...
+const ResourcesList = ({themeStr, sheetsCount, setConnectionsMode, topicsCount}) => {
+  return (
+      <ConnectionsPanelSection title={strings.resources}>
+        <ToolsButton
+          text={strings.sheets}
+          icon={iconData.get('sheet', themeStr)}
+          count={sheetsCount}
+          onPress={()=>{ setConnectionsMode("sheetsByRef"); }}
+        />
+        {!topicsCount ? null : <ToolsButton
```
nsantacruz, line 447: switch the order of this if statement so it is positive (not using !). This tends to be simpler to understand.

ConnectionsPanel.js
YishaiGlasner: const { theme, interfaceLanguage } = useContext(GlobalStateContext);
theme.tertiaryText
throws error
this is way to get theme is also the usage wat in ToolsButton

ConnectionsPanel.js
YishaiGlasner: i'm not sure i understand the question. i see here: 1. external View. 2. internal optional View for the title (with different styles). 3. internal View for the children

ConnectionsPanel.js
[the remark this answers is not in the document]
YishaiGlasner: i don't know. this is something from the outer code that i didn't taks. mayve this shouldn't be here, but i think it's not in this scope.

ConnectionsPanel.js
nsantacruz: You're right. I forgot there's a hook in Hooks.js called `useGlobalState`. This can be used like this: 

```javascript
const { theme, interfaceLanguage } = useGlobalState();
```

ConnectionsPanel.js
```diff
@@ -385,11 +385,11 @@ ToolsButton.propTypes = {
 
 const ConnectionsPanelSection = ({ title, children }) => {
   const { themeStr, interfaceLanguage } = useContext(GlobalStateContext);
-  const theme = getTheme(themeStr);
+  const theme= getTheme(themeStr);
   return (
     <View style={styles.connectionPanelSection} >
       {title &&
-        <View style={styles.connectionPanelTitle}>
+        <View style={{...styles.connectionPanelTitle, ...theme.lightGreyBorder}}>
```
nsantacruz, line 437: use ```style={[styles.connectionPanelTitle, theme.lightGreyBorder]}```. This is the standard way to combine multiple themes.

ConnectionsPanel.js
nsantacruz: Yeah, I got confused.

nsantacruz (approved):

PR #112, opened 2024-03-06
Translations

YishaiGlasner:
Make version choosing more compatible to web:
0. functions for offline and online getting translations from api-v3.
1. In the AboutBox: show alternate source versions, show versions details similar web.
2. In the translations box: show only translations, show the content of the translation rather than its versionTitle, show versions details similar web - that's done with a new component VersionBlockWithPreview.
3. In VersionBlock.js: New component VersionBlockHeader, VersionMetadata, and LinkWithKey for the use of both old VersionBlock and new VersionBlockWithPreview. New functions useVersionTitle and useTextAlign.
4. Refactor openActionSheet and tempOpenRef from TextList to Misc.js so VersionBlock can use it.
5. Garbage cleanup.


offline.js
```diff
@@ -22,6 +22,30 @@ export const loadTextOffline = async function(ref, context, versions, fallbackOn
     return {result: textFromRefData(processed)};
 };
 
+export const getAllTranslations = async function (ref, versions) {
```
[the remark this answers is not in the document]
nsantacruz: Also, `versions` doesn't seem to make sense as a parameter here. If I understand correctly, it's used to avoid a call to `getOfflineVersionObjectsAvailable` which doesn't seem to be a very expensive function. By passing in `versions` it makes the control flow confusing. Is versions an internal variable or is it an input?

offlineOnline.js
```diff
@@ -51,6 +51,20 @@ export const loadVersions = async (ref) => {
     return { versions, versionsApiError };
 };
 
+export const loadTranslations = async (ref, versions) => {
+    let translationsApiError = false;
```
nsantacruz, line 55: I see that you copied this from the function I wrote above (`loadVersions`). I'm now reading this and realizing how terrible this pattern is. We should not be storing a boolean for whether or not there was an error. We should throw the error and have the outer scope catch it. Happy to huddle to explain.

api.js
```diff
@@ -174,6 +174,10 @@ var Api = {
             urlSuffix += `&stripItags=1`;
           }
           break;
+        case 'translations':
```
nsantacruz, line 177: I think we should expose the full text v3 api and allow parameters to be passed to specify what to return as opposed to only allowing translations.

TranslationsBox.js
```diff
@@ -64,43 +63,47 @@ const useVLangState = (currVersionObjects, versions) => {
...
-    setVLangState
-  } = useVLangState(currVersionObjects, versions);
+  const {theme, interfaceLanguage} = useGlobalState();
+  const {vLangState, setVLangState } = useVLangState(
+      currVersionObjects,
+      []
+  );
+  const [apiError, setApiError] = useState(false)
+  const loadData = async () => {
+    setApiError(false);
```
nsantacruz, line 81: why are these two lines necessary? presumably the values will be set properly once the api call finishes

VersionBlock.js
```diff
@@ -1,108 +1,207 @@
...
-    versionTitle = version['versionTitle'];
-    versionSource = version['versionSource'];
-    shortVersionSource = Sefaria.util.parseURLhost(versionSource);
-    license = version['license'];
-    licenseURL = Sefaria.util.getLicenseURL(license);
-    versionNotes = version['versionNotes'];
-  }
+  const {theme} = useGlobalState();
+  const innerText = (
+    <SText lang={"english"} style={[styles.en, styles.textTocVersionTitle, useTextAlign(), theme.text]}>
```
nsantacruz, line 45: it's dangerous to put a hook as a style because there may be a conditional added before this line that returns early. Better to define it at the top of the component.

nsantacruz: Or change the function to be `getTextAlign(interfaceLanguage)` so it's not a hook

VersionBlock.js
```diff
@@ -1,108 +1,207 @@
...
+      {useVersionTitle(version)}
```
nsantacruz, line 46: Same here

offline.js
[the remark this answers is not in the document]
YishaiGlasner: i'm adding offline for we have it in all functions, but in principle i think we should remove all the offline from functions in module that named offline, if there is no a good reason.

TranslationsBox.js
```diff
@@ -64,43 +63,47 @@ const useVLangState = (currVersionObjects, versions) => {
...
 }) => {
-  const { themeStr, interfaceLanguage } = useContext(GlobalStateContext);
-  const {
-    vLangState,
-    setVLangState
-  } = useVLangState(currVersionObjects, versions);
+  const {theme, interfaceLanguage} = useGlobalState();
+  const {vLangState, setVLangState } = useVLangState(
+      currVersionObjects,
+      []
```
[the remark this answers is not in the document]
YishaiGlasner: it's a different object. the versions prop don't have the text. anyway, i'm removing this prop (it was used only for the offline getAllTranslations, where it should be removed).

TranslationsBox.js
YishaiGlasner: when moving from segment to segment, i think we don't want the translation of the previous segment to remain until we have the response.

VersionBlock.js
YishaiGlasner: the idea is not to repat the code that creates the style from interfaceLanguage, and getting interfaceLanguage in any component. is there a way to do this?

Misc.js
```diff
@@ -1542,6 +1543,26 @@ const GreyBoxFrame = ({ children }) => {
   );
 };
 
+const openActionSheet = (refStr, versions, openRef, interfaceLanguage, heRefStr, category) => {
```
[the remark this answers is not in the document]
YishaiGlasner: honestly, i only moved this function for using it. i don't know how it should behave in any case.

nsantacruz (approved):

PR #127, opened 2024-07-15
End of book

YishaiGlasner:
bugfix: enables showing the end of a book when the connections panel is open.


YishaiGlasner:
will push another solution in another branch

PR #138, opened 2024-11-03
fix(search): use getFilter for deciding whether to update aggregation…

YishaiGlasner:
fixes 2 bugs:
1. search > filter > another search. we need another search for the aggregations. for that the aggregations update refactore out to updateAggregations, and used again in this case.
2. search > filter > click on result > back. here we want the search to return without aggregations (so it will do another search for them). for that the search prop for that is now based on getFilters rather than filtersValid.


ReaderApp.js
```diff
@@ -1620,11 +1620,28 @@ class ReaderApp extends React.PureComponent {
     });
   };
   onQueryChange = (type, query, resetQuery, fromBackButton, getFilters) => {
+
+    const updateAggregations = (data) => {
```
nsantacruz, line 1624: I think this function should be in the outer scope because it's so large. It's hard to know what dependencies it has on the variables in `onQueryChange` which make it difficult to change.

ReaderApp.js
```diff
@@ -1686,19 +1703,12 @@ class ReaderApp extends React.PureComponent {
...
-                const { availableFilters: tempAvailable, registry: tempRegistry, orphans: tempOrphans } = Sefaria.search[build_and_apply_filters](buckets, appliedFilters, appliedFilterAggTypes, aggregation, Sefaria);
-                availableFilters.push(...tempAvailable);  // array concat
-                registry = {...registry, ...tempRegistry};
-                orphans.push(...tempOrphans);
-                this.setAvailableSearchFilters(type, availableFilters, orphans);
-              }
-            }
+            updateAggregations(data);
+          } else if (getFilters) {
+            [queryProps.appliedFilterAggTypes, queryProps.applied_filters, queryProps.aggregationsToUpdate] = [[], [], ['path']];
```
nsantacruz, line 1708: why is aggregationsToUpdate hard-coded to 'path'? I don't believe this works for sheet search because there are different filters there.

nsantacruz (dismissed):

nsantacruz (approved):

PR #158, opened 2025-04-02
implement text reference fallback mechanism

schreibersefaria:
fix: implement text reference fallback mechanism

- Enhance refUpOne function to support slicing on spaces to go up to book level
- Add recursive fallback to try parent references when text isn't found
- Add failSilently flag to prevent multiple error popups during recursive calls
- Fix incorrect variable name (textUnavailableFromWebMessage)
- Add minimal logging for better debugging
- Remove small amount of unused code

This change improves user experience by attempting to find parent references when a specific text reference is unavailable. There are different part of the code (both on mobile and server) that deal with this problem in different situations. We thought of solving this in the server or trying to understand what causes the problem but decided to stay in the scope of the ticket.

Refs: Fail Gracefully When Navigating to a Specific [sc-33279]


ReaderApp.js
```diff
@@ -733,9 +737,22 @@ class ReaderApp extends React.PureComponent {
 
             resolve();
         }).catch(error => {
-          console.log(error);
+          console.log(`Dealing with error: ${error}. Ref: ${ref}`);
           if (error == "Return to Nav") {
-            this.openTextTocDirectly(Sefaria.textTitleForRef(ref));
+            // In case of unfound references, try going one ref up (up to the book) before dealing with error by returning to nav.
+            // We do this to avoid failing in case of references to a non existent ref after a changing index of in book.
```
YishaiGlasner, line 743: typo in "of in book"

ReaderApp.js
```diff
@@ -733,9 +737,22 @@ class ReaderApp extends React.PureComponent {
 
             resolve();
         }).catch(error => {
-          console.log(error);
+          console.log(`Dealing with error: ${error}. Ref: ${ref}`);
           if (error == "Return to Nav") {
-            this.openTextTocDirectly(Sefaria.textTitleForRef(ref));
+            // In case of unfound references, try going one ref up (up to the book) before dealing with error by returning to nav.
```
YishaiGlasner, line 742: by "references" do you mean ref? we call them ref; calling them reference is confusing

ReaderApp.js
```diff
@@ -733,9 +737,22 @@ class ReaderApp extends React.PureComponent {
...
+            // We do this to avoid failing in case of references to a non existent ref after a changing index of in book.
+            const refUpOne = Sefaria.refUpOne(ref, true);
+            // Break if there is no more ref up to do.
+            // Number limit is just in case of a edge case.
+            // refUpOne checks if book exists, so code wont go into this if if the book doesn't exist
+            if (ref !== refUpOne && numTries <= 4) {
```
YishaiGlasner, line 748: i think we're using numTries for cases of low connectivity. i don't think it's a good idea to use it also for number of going up in ref. i think a while loop will be better here.
also, i can't see how it works for if `error == "Return to Nav"`, it seems that `numTries` must be 5.

sefaria.js
```diff
@@ -83,9 +83,26 @@ Sefaria = {
...
-    return ref.lastIndexOf(":") !== -1 ? ref.slice(0, ref.lastIndexOf(":")) : ref;
+  refUpOne: function (ref, splitOnSpaces = false) {
+    /**
+     * return ref up one level, assuming you can
+     * 
+     * @param {str} ref - reference to go one up on
+     * @param {bool} split_on_spaces - Continue spliting on spaces until you reach a valid book. Doesn't split beyond that (Words in a book)
+     */
+
+    if (ref.lastIndexOf(":") !== -1) {
```
YishaiGlasner, line 94: `const lastIndexOfColon = ref.lastIndexOf(":");`
will prevent duplication and improve readabilty (this true also for lastIndexOfSpace).

sefaria.js
```diff
@@ -83,9 +83,26 @@ Sefaria = {
...
+      let newRef = ref.slice(0, ref.lastIndexOf(":"));
+      return newRef;
+    }
+    // Only return sliced ref on space if the new ref will have a valid title in it
+    else if (splitOnSpaces && ref.lastIndexOf(" ") !== -1) {
```
YishaiGlasner, line 99: i think it will be cleaner to do it in another function and call it here.

sefaria.js
```diff
@@ -83,9 +83,26 @@ Sefaria = {
...
+    /**
+     * return ref up one level, assuming you can
+     * 
+     * @param {str} ref - reference to go one up on
+     * @param {bool} split_on_spaces - Continue spliting on spaces until you reach a valid book. Doesn't split beyond that (Words in a book)
+     */
+
+    if (ref.lastIndexOf(":") !== -1) {
+      let newRef = ref.slice(0, ref.lastIndexOf(":"));
+      return newRef;
```
YishaiGlasner, line 96: i think it will be better to do
let newRef = ref;
in the beginning, and on `return newRef` statement in the end.

YishaiGlasner:
maybe it'd be better to take the logic to the server (that would suggest a ref and the clinet would make another request with it):
1. the server has much more information about potential refs.
2. it will save duplication of the logics in web.
3. will help also third parties using our api.

ReaderApp.js
schreibersefaria: I couldn't think of a solution if we get a ref that looks like this Genesis 999.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1
Other than something like that, the recursive call should deal with the refUpOne. I'm not sure I get why you suggest a while?
Also, why would it have to be 5?
Maybe a huddle would be easiest for this.

sefaria.js
schreibersefaria: I generally refactored the func, could you pls have another look?

sefaria.js
schreibersefaria: Would a ref at this point in the code always be <Book level1:level2:etc>?
Trying to think what the new func should be (was thinking to call it extractBook, if that is the format)

sefaria.js
schreibersefaria: Mmm but wouldn't that be less readable. As in, now it's clear the function returns an unchanged ref if nothing was changed?

sefaria.js
schreibersefaria: We agreed on a solution on a call

ReaderApp.js
schreibersefaria: We agreed on a solution on a call. I removed the numTries assertion

YishaiGlasner (dismissed):

yitzhakc (approved):

YishaiGlasner (approved):

PR #160, opened 2025-04-29
Fix(_toURL): Empty Version Objects Causing URL Construction Errors

schreibersefaria:
## Issue
Bookmarks created from topic pages were causing an endless loading spinner, with the error:

    url.replace is not a function

When navigating from a topic page to a text and bookmarking it, the app was storing version info as empty objects (`{ en: {}, he: {} }`) instead of strings or `null`.

## Fix
Verify that each version value is actually a string before sanitizing it, so that we never call `.replace()` on an object.

## Implementation
In **`api.js`** (_toURL method), wrap the version-suffix logic in a `typeof` check:

    // Before sanitizing, ensure versions.en is a string
    if (versions.en && typeof versions.en === 'string') {
      urlSuffix += `&ven=${this._sanitizeURL(versions.en)}`;
    }

    // Similarly for Hebrew, if needed:
    if (versions.he && typeof versions.he === 'string') {
      urlSuffix += `&vhe=${this._sanitizeURL(versions.he)}`;
    }

## Next Steps
1. **Verify version data**  
   What should happen when those version-objects actually contain valid version strings?  
2. **E2E test**  
   Ensure bookmarking texts from topic pages now works and does not break existing flows.
   The bookmark that is saved can also not be deleted.

### DIFF
--- a/api.js
+++ b/api.js
@@ -171,9 +171,6 @@ var Api = {
           url += 'api/texts/';
           urlSuffix = `?context=${context === true ? 1 : 0}&commentary=0`;
           if (versions) {
-            // BUGFIX Patch: Bookmarks created from topic can pass empty objects {en: {}, he: {}}
-            // instead of strings or nulls for version values, causing url.replace errors.
-            // We check that each version is actually a string before trying to sanitize it.
             if (typeof versions.en === 'string') {
               urlSuffix += `&ven=${this._sanitizeURL(versions.en)}`;
             }
--- a/api.js
+++ b/api.js
@@ -36,20 +36,6 @@ var Api = {
   _lexiconCache: {},
   _currentRequests: {}, // object to remember current request in order to abort. keyed by apiType
   
-  // Utility function to sanitize version objects
-  sanitizeVersions: function(versions) {
-    if (!versions) return versions;
-    
-    const sanitized = {...versions};
-    if (sanitized.en && typeof sanitized.en === 'object' && Object.keys(sanitized.en).length === 0) {
-      delete sanitized.en;
-    }
-    if (sanitized.he && typeof sanitized.he === 'object' && Object.keys(sanitized.he).length === 0) {
-      delete sanitized.he;
-    }
-    return sanitized;
-  },
-  
   _textCacheKey: function(ref, context, versions) {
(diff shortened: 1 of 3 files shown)


api.js
```diff
@@ -170,8 +171,15 @@ var Api = {
           url += 'api/texts/';
           urlSuffix = `?context=${context === true ? 1 : 0}&commentary=0`;
           if (versions) {
-            if (versions.en) { urlSuffix += `&ven=${this._sanitizeURL(versions.en)}`; }
-            if (versions.he) { urlSuffix += `&vhe=${this._sanitizeURL(versions.he)}`; }
+            // BUGFIX Patch: Bookmarks created from topic can pass empty objects {en: {}, he: {}}
+            // instead of strings or nulls for version values, causing url.replace errors.
+            // We check that each version is actually a string before trying to sanitize it.
+            if (versions.en && typeof versions.en === 'string') {
```
YishaiGlasner, line 177: you don't need the `versions.en` anymore
it can be just
`typeof versions.en === 'string'`
if it's an empty string, adding it will do no harm (unless this runs so many so it has performance effect).
also in l30

api.js
```diff
@@ -170,8 +171,15 @@ var Api = {
           url += 'api/texts/';
           urlSuffix = `?context=${context === true ? 1 : 0}&commentary=0`;
           if (versions) {
-            if (versions.en) { urlSuffix += `&ven=${this._sanitizeURL(versions.en)}`; }
-            if (versions.he) { urlSuffix += `&vhe=${this._sanitizeURL(versions.he)}`; }
+            // BUGFIX Patch: Bookmarks created from topic can pass empty objects {en: {}, he: {}}
```
YishaiGlasner, line 174: i think this doc string is redundant.
checking that something is string before adding it to a string is self evident.

api.js
schreibersefaria: Done. is that true also for HE?

api.js
schreibersefaria: removed

schreibersefaria: Although maybe it makes sense to at least explain the behaviour. Why we chose to ignore non strings instead of throwing an error?

YishaiGlasner (approved):

PR #163, opened 2025-05-05
chore: Build files from fastlane

schreibersefaria:
I forgot to add these to the version branch, so here they are.


schreibersefaria:
This will likely be irrelevant after the move to RN 77.
@nsantacruz Is that right? Once you do the next build, these aren't helpful?

PR #165, opened 2025-05-13
rn-77-fix-toc-arrows

nsantacruz:
The arrows on complex text table of contents had incorrect spacing after upgrading to RN 77. This PR fixes that.

### DIFF
--- a/Misc.js
+++ b/Misc.js
@@ -725,11 +725,17 @@ const CollapseIcon = ({ showHebrew, isVisible }) => {
     }
   }
   return (
-    <Image
-      source={src}
-      style={[(showHebrew ? styles.collapseArrowHe : styles.collapseArrowEn), Platform.OS === 'android' ? {marginTop: 3} : null]}
-      resizeMode={'contain'}
-    />
+      // arrow margins vary greatly depending on the platform and language
+      <View style={[
+          (showHebrew ? styles.collapseArrowOuterHe : styles.collapseArrowOuterEn),
+          (Platform.OS === 'android' ? (showHebrew ? styles.collapseArrowAndroidHe : styles.collapseArrowAndroidEn) : (showHebrew ? styles.collapseArrowIOSHe : null))
+      ]}>
+        <Image
+            source={src}
+            style={styles.collapseArrow}
+            resizeMode={'contain'}
+        />
+      </View>
   );
 }
 CollapseIcon.propTypes = {
--- a/Misc.js
+++ b/Misc.js
@@ -725,10 +725,12 @@ const CollapseIcon = ({ showHebrew, isVisible }) => {
     }
   }
   return (
-      // arrow margins vary greatly depending on the platform and language
+      // React Native 0.77 caused differences in spacing for this arrow icon on iOS and Android and English and Hebrew
+      // We are accounting for this with fine-grained styling below
       <View style={[
           (showHebrew ? styles.collapseArrowOuterHe : styles.collapseArrowOuterEn),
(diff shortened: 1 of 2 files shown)


ReaderApp.js
```diff
@@ -25,7 +25,7 @@ import BackgroundFetch from "react-native-background-fetch";
 import { InAppBrowser } from 'react-native-inappbrowser-reborn';
 import BootSplash from "react-native-bootsplash";
 import nextFrame from 'next-frame';
-import RNShake from 'react-native-shake';
+import { Shake } from '@sapkalabs/react-native-shake';
```
schreibersefaria, line 28: This seems flaky. It's a library from 2 months ago and could easily not be maintained.

android/gradle.properties
```diff
@@ -33,7 +33,7 @@ reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 # your application. You should enable this flag either if you want
 # to write custom TurboModules/Fabric components OR use libraries that
 # are providing them.
-newArchEnabled=false
+newArchEnabled=true
```
schreibersefaria, line 36: This seems like a big change (and I think there are some others), maybe worth adding them to the PR description

schreibersefaria (commented):
This is an outdated review

useHTMLViewStyles.js
```diff
@@ -12,7 +12,7 @@ export function useHTMLViewStyles(bilingual, textType) {
       const isHeb = textType === "hebrew";
       const lineHeightMultiplier = isHeb ? (Platform.OS === 'android' ? 1.333 : 1.25) : 1.15;
       const fontSizeMultiplier = isHeb ? 1 : 0.8;
-      const justifyStyle = {textAlign: (isStacked && Platform.OS === 'ios') ? 'justify' : (isHeb ? 'right' : 'left'), writingDirection: isHeb ? "rtl" : "ltr"};
+      const justifyStyle = {textAlign: isStacked ? 'justify' : (isHeb ? 'right' : 'left'), direction: isHeb ? "rtl" : "ltr"};
```
schreibersefaria: Personally I would go for consistency rather than allowing justify when we can. So no justify at all. Sounds like a Micky question.
In any case, I would add a comment with the current behaviour we are going for

Misc.js
```diff
@@ -725,11 +725,17 @@ const CollapseIcon = ({ showHebrew, isVisible }) => {
     }
   }
   return (
-    <Image
-      source={src}
-      style={[(showHebrew ? styles.collapseArrowHe : styles.collapseArrowEn), Platform.OS === 'android' ? {marginTop: 3} : null]}
-      resizeMode={'contain'}
-    />
+      // arrow margins vary greatly depending on the platform and language
```
schreibersefaria, line 728: It may be worth adding long form comments with the current state of what dependencies we know about and are taking into account (whenever we have styling based on OS and lang) so we can more easily fix it later.
(This is also true for justification etc)

Misc.js
```diff
@@ -725,11 +725,17 @@ const CollapseIcon = ({ showHebrew, isVisible }) => {
...
+      <View style={[
+          (showHebrew ? styles.collapseArrowOuterHe : styles.collapseArrowOuterEn),
+          (Platform.OS === 'android' ? (showHebrew ? styles.collapseArrowAndroidHe : styles.collapseArrowAndroidEn) : (showHebrew ? styles.collapseArrowIOSHe : null))
```
schreibersefaria, line 731: could use indentation

PR #166, opened 2025-05-15
Bug/sc 34121/ios app v 3 4 31 in testflight reference jastrow

schreibersefaria:
The problem was being caused by a race condition of updateDataNext and updating the state for the previous section by _loadRelatedOnlineAndOffline. The problem accord because we were not updating the state in an immutable way.

This PR focuses on updating the state updated ReaderApp to be more like the recommended react state update.

### DIFF
--- a/ReaderApp.js
+++ b/ReaderApp.js
@@ -906,11 +906,7 @@ class ReaderApp extends React.PureComponent {
     } catch(error) {
       versionsApiError = true;
     }
-    this.setState(prevState => ({ 
-      ...prevState,
-      versions, 
-      versionsApiError 
-    }));
+    this.setState({versions, versionsApiError});
   };
 
   updateData = (direction) => {
--- a/ReaderApp.js
+++ b/ReaderApp.js
@@ -476,15 +476,18 @@ class ReaderApp extends React.PureComponent {
   setTextLanguage = (textLanguage, textFlow) => {
     // try to be less dependent on state in this func because it is called in componentWillUpdate
     textFlow = textFlow || this.state.textFlow;
+    
     this.props.dispatch({
       type: STATE_ACTIONS.setTextLanguage,
       value: textLanguage,
     });
+    
     if (textLanguage === "bilingual" && textFlow === "continuous") {
       this.setTextFlow("segmented");
     }
+    
     this.closeReaderDisplayOptionsMenu();
-  }
+  };
 
   setTheme = themeStr => {
     this.props.dispatch({
@@ -492,7 +495,7 @@ class ReaderApp extends React.PureComponent {
       value: themeStr,
     });
     this.closeReaderDisplayOptionsMenu();
-  }
+  };
 
   setAliyot = show => {
     this.props.dispatch({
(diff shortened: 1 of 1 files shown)


ReaderApp.js
```diff
@@ -1062,22 +1093,28 @@ class ReaderApp extends React.PureComponent {
...
+      sectionArray: [],
+      sectionHeArray: [],
+      offsetRef: null,
+      connectionsMode: null,
+    }), () => {
       this.closeMenu(); // Don't close until these values are in state, so sheet can load
-      this.setState({
+      
+      // Second setState
+      this.setState(prevState => ({
```
nsantacruz, line 1111: simple setState should work since new state isn't dependent on old state

ReaderApp.js
```diff
@@ -1442,13 +1479,14 @@ class ReaderApp extends React.PureComponent {
       }
       const nextFilter = new LinkFilter(name, heName, collectiveTitle, heCollectiveTitle, nextRefList, nextHeRefList, category);
 
-      this.state.linkRecentFilters[filterIndex] = nextFilter;
+      const newLinkRecentFilters = [...this.state.linkRecentFilters];
```
nsantacruz, line 1482: wrap this in setState callback and use prevState when calculating `newLinkRecentFilters`

ReaderApp.js
```diff
@@ -1499,9 +1537,15 @@ class ReaderApp extends React.PureComponent {
...
     }
-    const newLinkContents = [...this.state.linkContents];
-    newLinkContents[pos] = data;
-    this.setState({linkContents: newLinkContents});
+    
+    this.setState(prevState => {
+      const newLinkContents = [...prevState.linkContents];
+      newLinkContents[pos] = data;
+      return {
+        ...prevState,
```
nsantacruz, line 1545: no need to pass prevState here

ReaderApp.js
```diff
@@ -1517,12 +1561,18 @@ class ReaderApp extends React.PureComponent {
       filterIndex = this.state.versionFilterIndex;
     }
     if (!segmentRef) { segmentRef = this.state.segmentRef; }
-    this.state.versionRecentFilters[filterIndex].refList = [segmentRef];
+    
+    const newVersionRecentFilters = [...this.state.versionRecentFilters];
```
nsantacruz, line 1565: wrap this logic in setState callback and calculate `newVersionRecentFilters` using prevState

ReaderApp.js
```diff
@@ -1532,8 +1582,11 @@ class ReaderApp extends React.PureComponent {
       // only want to show versionLanguage in results
       const removeLang = versionLanguage === "he" ? "en" : "he";
       data.result[removeLang] = "";
-      this.state.versionContents[pos] = data.result;
-      this.setState({versionContents: this.state.versionContents.slice(0)});
+      
+      const newVersionContents = [...this.state.versionContents];
```
nsantacruz, line 1586: wrap this logic in setState callack and calculate `newVersionContents` with prevState

ReaderApp.js
```diff
@@ -1597,10 +1650,14 @@ class ReaderApp extends React.PureComponent {
   onTextListDragEnd = evt => {
     const headerHeight = 75;
     const flex = 1.0 - (evt.nativeEvent.pageY-headerHeight)/(ViewPort.height-headerHeight) + this._textListDragOffset;
+    
     if (flex > 0.9 || flex < 0.2) {
       this.animateTextList(flex, flex > 0.9 ? 0.9999 : 0.0001, 200);
     } else {
-      this.setState({ textListFlexPreference: flex });
+      this.setState(prevState => ({
```
nsantacruz, line 1657: simple setState should work here since there's no dependence on prevState

ReaderApp.js
```diff
@@ -604,61 +607,69 @@ class ReaderApp extends React.PureComponent {
...
+          this.animateTextList(0.0001, this.state.textListFlexPreference, 200);
         });
       }
-      this.forceUpdate();
+      Sefaria.history.saveHistoryItem(this.getHistoryObject, true);
+    });
+    
+    // NOTE: This forceUpdate is potentially problematic - it forces a re-render
+    // immediately after setting state, which can lead to inconsistencies
+    this.forceUpdate();
```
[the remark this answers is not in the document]
schreibersefaria: Well I left a note without changing because I don't know what the function does.
From a short test nothing happens but I may not be even triggering the func so I think it's best kept out of the task scope / We can get on a quick huddle to check together.

schreibersefaria:
@nsantacruz thanks for all the comments.
Next time I will try to do another step before a PR if I don't fully understand the solution. Get on a call with you or something.

nsantacruz (approved):

PR #167, opened 2025-05-19
Update proptypes to deprecated lib

schreibersefaria:
Changed all the calls to the deprecated prop-types to now use, the aptly named, community supported, deprecated-react-native-prop-types.
Removed the old import from packages


schreibersefaria:
Looks like this is the wrong direction

PR #169, opened 2025-05-26
Fix crashing

YishaiGlasner:
## Background
1. It seems that RN have changed the way of handling setState, and that caused a bug of infinite loading.
2. We have in our code a bad patterm of changing the state out of setState and using setState.
3. Daniel solved this, by changing the bad patterm in one place. Also he he changed the bad patterm in some other places.
4. The problem is that our code was built on this bad patterm, so it sometimes assume that the state change is immediate.

## Solution
I have reverted the changes except the one that sloved the infinite loading, and it seems to work.

## Note
It will be great to change the manually state setting, but it should be done carefully.


schreibersefaria (commented):
There is something confusing to me here.
I did many more changes than this.
https://github.com/Sefaria/Sefaria-Mobile/pull/166/files
Did you revert only the changes we talked about specifically?

schreibersefaria (approved):

yitzhakc (approved):

PR #181, opened 2025-08-10
fix-deep-linking-missing-native-code

nsantacruz:
When we upgraded our AppDelegate from Objective C to Swift we lost some boiler-plate deep linking code that React Native needs to support deep linking in the app. This PR adds it in again.

Note, counter-intuitively deep links still worked when the app was launched through a deep link. This is because these deep links are handled only in JS. However, for deep links while the app is running, they require native code to be handled.


schreibersefaria (approved):
Can't say I understood much of this but this is just returning code we have so if it works, great! Seems quite boiler-plate as you put it.

PR #190, opened 2025-12-30
Feature/sc 19749/add works on sefaria to author pages in mobile

dcschreiber:
This is not going to be merged before modularization launches. 

Adds "Works on Sefaria" section to author topic pages like in the Web version.
Before: Author pages showed a "Sources" tab with text references mentioning the author.

  Changes

api.js
  - Add with_indexes parameter to topic API call to fetch author indexes

LocalizedStrings.js
  - Add localized string: "Works on Sefaria" / "חיבורים וכתבים בספריא"

TopicPage.js
  - Detect author topics via subclass === 'author'
  - Show "Works on Sefaria" tab (instead of "Sources") when author has indexes
  - Add AuthorWork component using CategoryButton for consistent styling with topic pages
  - Hide search bar for works tab
  - Tapping a work navigates in-app using handleOpenURL


TopicPage.js
```diff
@@ -327,8 +327,14 @@ const TopicPage = ({ topic, onBack, openNav, openTopic, showToast, openRef, setT
   const [parashaData, setParashaData] = useState(null);
   const [query, setQuery] = useState(null);
   const [portal, setPortal] = useState(null);
+  const isAuthor = topicData?.subclass === 'author';
+  const hasWorks = isAuthor && topicData?.indexes?.length;
   const tabs = [];
-  if (!!topicData && !!topicData.textRefs.length) { tabs.push({text: strings.sources, id: 'sources'}); }
+  if (hasWorks) {
```
nsantacruz, line 334: now that sheets are out of the app, it seems odd to keep using tabs when there's only one tab.

dcschreiber: I agree but think that may be a bit out of scope for this task. 
Mickey actually coincidently mentioned he is OK with the tab style for this change

nsantacruz (dismissed):

nsantacruz (approved):

PR #196, opened 2026-03-29
chore: builds and runs with ios 26 sdk

stevekaplan123:
## Description

This PR updates the mobile app so the iOS project builds with the iOS 26 SDK / newer Xcode toolchain.

The main change is an iOS dependency and build-system refresh: React Native iOS pods move from `0.79.5` to `0.79.7`, CocoaPods now builds Firebase-related pods as static frameworks, and Expo’s root-view customization is patched so `RNBootSplash` still initializes correctly with the New Architecture path.

## Code Changes

- Updated the app version from `6.6.0` to `6.7.0`.
- Updated Android build metadata to `VERSION_BUILD=62061037` and `VERSION_MINOR=7`.
- Updated iOS React Native pods from `0.79.5` to `0.79.7`.
- Changed the iOS CocoaPods setup to use static frameworks by default, which is needed for Firebase Swift pods.
- Added a post-install patch for `fmt` so newer Apple Clang versions do not fail on `consteval` format-string checks.
- Moved Firebase JS initialization earlier in app startup so Firebase namespaces are registered before analytics/crashlytics calls.
- Replaced the old Expo root-view customization hook with a patched New Architecture-safe hook used by `RNBootSplash`.
- Refreshed bundled source JSON files for iOS and Android.
- Replaced the old React Native patch with a new Expo `53.0.20` patch.

Most of the large diff is generated iOS project / CocoaPods output. The important hand-written changes are in:

- `ios/Podfile`
- `ios/ReaderApp/AppDelegate.swift`
- `index.js`
- `ReaderApp.js`
- `patches/expo+53.0.20.patch`

`package-lock.json` changed, but package versions did not change there; the lockfile mostly updates npm metadata flags.


patches/expo+53.0.20.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/node_modules/expo/ios/AppDelegates/ExpoReactNativeFactory.swift b/node_modules/expo/ios/AppDelegates/ExpoReactNativeFactory.swift
```
yitzhakc, line 1: I'm assuming we don't need this file in Git

yitzhakc (changes requested):

patches/expo+53.0.20.patch
stevekaplan123: I believe we need the patches.  The packages "have to be in git. patch-package reads them at install time, so without them in git, every fresh yarn install (CI, new dev machine, etc.) ships node_modules without the fixes and you get the Android crash / iOS build failure back."

PR #199, opened 2026-04-27
Bump rn to 81

stevekaplan123:
## Description

This PR upgrades the app from **React Native 0.79 to 0.81**, along with a bunch of related dependency and build-tool bumps that the upgrade requires. It also switches the project's package manager back to **Yarn** (from npm), which turned out to be necessary to make the 0.81 build actually work.

### 1. Core version bumps in `package.json`
- `react-native`: 0.79 → **0.81.6**
- `react`: 19.0.0 → **19.1.4**
- `expo`: 53 → **54**, `expo-file-system`: 18 → **19**
- RN community CLI packages: 18 → **20**
- `@react-native/babel-preset`, `eslint-config`, `metro-config`: 0.79 → **0.81**
- Various RN libraries (gesture-handler, safe-area-context, svg, webview, async-storage, netinfo) bumped to versions compatible with RN 0.81.
- Added `htmlparser2` (now needs to be a direct dep instead of pulled in transitively).

### 2. `expo-file-system` API change
Expo 54 rewrote the file system module. The old API is still available but only under a `/legacy` import path. To avoid rewriting all our download/offline logic right now, every `import * as FileSystem from 'expo-file-system'` was changed to `'expo-file-system/legacy'` in `DownloadControl.js`, `SettingsPage.js`, `history.js`, `offline.js`, and `sefaria.js`. Functionally identical; just keeps the old API.

### 3. Android build updates
- `compileSdk` / `targetSdk`: 35 → **36** (Android 16 SDK), with matching `buildToolsVersion`.
- Kotlin: 2.0.21 → **2.1.20**.
- Gradle wrapper: 8.13 → **8.14.3**.
- New flag in `android/gradle.properties`: `edgeToEdgeEnabled=false`. RN 0.81 introduced an "edge-to-edge" display mode (drawing under the status/nav bars). We're explicitly opting out for now to avoid layout regressions.
- `android/app/src/main/java/org/sefaria/sefaria/MainApplication.kt`: RN 0.81 simplified app startup. We replaced the manual `SoLoader.init(...)` + `DefaultNewArchitectureEntryPoint.load()` calls with a single `loadReactNative(this)` call, which is the new official entry point.

### 4. iOS build updates
- `ios/ReaderApp/Info.plist`: added `RCTNewArchEnabled = true`. RN 0.81 reads the New Architecture flag from `Info.plist` on iOS instead of relying on a build-time define.
- `ios/ReaderApp.xcodeproj/project.pbxproj`: regenerated by `pod install` — minor input/output path changes for build script phases. Not hand-edited.
- Massive `ios/Podfile.lock` churn — also just the result of `pod install` with the new RN/Expo versions.

### 5. New patch file: `patches/react-native+0.81.6.patch`
We use `patch-package` to apply small fixes on top of node_modules. The old patch (`react-native+0.77.2.patch`) was removed and replaced with a 0.81.6 equivalent. It bumps the `fmt` C++ library (a formatting library that RN's iOS build pulls in) from 11.0.2 to 12.1.0 to resolve a CocoaPods dependency conflict.

### 6. Tiny JS fix in `sefaria.js`
Changed `Sefaria = { ... }` to `global.Sefaria = { ... }`. The newer Metro bundler in RN 0.81 is stricter about implicit globals — assigning to an undeclared variable now errors. Explicitly attaching to `global` keeps the same behavior.


YishaiGlasner (commented):
although i did bump RN before, i can't say i understand what's going on here. if it's working it's fine for me, but new file with 16K lines seem to be very very weird

stevekaplan123:
> although i did bump RN before, i can't say i understand what's going on here. if it's working it's fine for me, but new file with 16K lines seem to be very very weird

You're right.  Deleted that file which didn't need to be there.

nsantacruz (approved):

PR #200, opened 2026-04-27
Rn 81 with copy bug fix

stevekaplan123:
The only real change in this PR is to add a function, htmlToTextCanonical, which cleans HTML entities and HTML fragments out of text that is copied from the app.  It works essentially the same as htmlToText in static/js/sefaria/util.js.


saengel (approved):
Approved, contingent on two things:
1. Are all of these JSON files meant to be here? What is the 1-line change in each one? Hard for me to actually see the diff in GH. 
2. Copilot left a comment about tests, I'm not sure if relevant in Sefaria-Mobile - leaving that to your judgement/understanding of best practices here.

yitzhakc (approved):

PR #207, opened 2026-06-07
Bug/sc 43538/add the extra empty space placeholder to the

stevekaplan123:
This PR addresses a problem that at the bottom of a panel, the text was not visible.  The solution is to add padding.  For both the reader and connections panel, we add space to the bottom.  The flow is as follows: Reader or ConnectionsPanel renders → onLayout runs and reports its real height → we save it as 'height' → the bottom padding gets computed as 70% of that height.  Because onLayout re-fires on resize, dragging the panel divider re-measures and the padding stays proportional to the panel.


TextList.js
```diff
@@ -47,9 +47,19 @@ class TextList extends React.Component {
     this.state = {
       dataSource,
       updateScrollPosKey: true,
+      // Height of the list's own viewport, measured via onLayout. Used to size the
+      // bottom padding relative to the ConnectionsPanel rather than the full window.
+      scrollViewHeight: 0,
     };
   }
 
+  _onScrollViewLayout = (e) => {
```
YishaiGlasner: it seems we have this func 3 times. what about a helper func:
```
function updateScrollViewHeight(e, scrollViewHeight) {
  const { height } = e.nativeEvent.layout;

  if (height && height !== scrollViewHeight) {
    return height;
  }

  return scrollViewHeight;
}
```
?

YishaiGlasner: maybe it's not a great idea - i'd suggest you to consult your agents and see if there is a nice solution for DRY. if not, i'll approve

stevekaplan123: Good idea.  Added a new function makeScrollViewLayoutHandler that is now called three times.

Styles.js
```diff
@@ -9,6 +9,22 @@ const isIOS = Platform.OS === 'ios';
...
+// one-line last segment to scroll up to the top of the panel when pressed, without
+// scrolling it out of view. Applied at runtime against each panel's measured height
+// (not the window height) so it stays correct inside the ConnectionsPanel split view.
+export const SCROLL_PADDING_BOTTOM_FRACTION = 0.7;
+
+// Builds an onLayout handler that records a scrolling panel's own viewport height
+// into `component.state.scrollViewHeight`, so bottom padding can be sized relative
+// to the panel rather than the window. Bind once in the constructor:
+//   this._onScrollViewLayout = makeScrollViewLayoutHandler(this);
+export const makeScrollViewLayoutHandler = (component) => (e) => {
```
YishaiGlasner: i think this is difficult to read. i'd something like
```
export function makeScrollViewLayoutHandler(component) {
  return function onScrollViewLayout(e) {
```

YishaiGlasner (approved):

PR #211, opened 2026-06-25
Bug/sc 34147/app internet connection unavailable pop up

stevekaplan123:
## Description
Users were seeing "No Internet Connection" pop ups when offline even though they had downloaded the complete library.  This was because we were trying to access translations that had not been downloaded in the complete library.  (Note: "Complete library" does not include all translations).  This PR adjusts how translations are loaded so the app avoids triggering the “no internet connection” blocking alert.  An export change was also done as part of this project so that a user who downloads "Complete Library" will now have all translations on his phone, and a user who selects to download "Talmud" has all Talmud translations on his phone.

**Changes:**
- Passes the app’s connectivity state into `offlineOnline.loadTranslations`.
- Updates translation loading to only hit the API when online.


stevekaplan123:
@YishaiGlasner I removed the "fail silently" code so that now if a translation fails to load it will notify the user.  However, this code still fixes the original bug whereby scrolling in offline mode causes an unnecessary pop-up alerting the user they're offline.  I avoid this bug simply by checking if the user is online/offline when loading translations.

YishaiGlasner:
> @YishaiGlasner I removed the "fail silently" code so that now if a translation fails to load it will notify the user. However, this code still fixes the original bug whereby scrolling in offline mode causes an unnecessary pop-up alerting the user they're offline. I avoid this bug simply by checking if the user is online/offline when loading translations.

@stevekaplan123 so now if an offline user will tap translation and it will fail, he will have no message?

stevekaplan123:
> > @YishaiGlasner I removed the "fail silently" code so that now if a translation fails to load it will notify the user. However, this code still fixes the original bug whereby scrolling in offline mode causes an unnecessary pop-up alerting the user they're offline. I avoid this bug simply by checking if the user is online/offline when loading translations.
> 
> @stevekaplan123 so now if an offline user will tap translation and it will fail, he will have no message?

Probably yes, but that's a rare case, isn't it?

YishaiGlasner:
> > > @YishaiGlasner I removed the "fail silently" code so that now if a translation fails to load it will notify the user. However, this code still fixes the original bug whereby scrolling in offline mode causes an unnecessary pop-up alerting the user they're offline. I avoid this bug simply by checking if the user is online/offline when loading translations.
> > 
> > 
> > @stevekaplan123 so now if an offline user will tap translation and it will fail, he will have no message?
> 
> Probably yes, but that's a rare case, isn't it?

maybe, but this is not an edge case - he's using translations in the way he should and we have this error exactly for such cases

stevekaplan123:
> > > > @YishaiGlasner I removed the "fail silently" code so that now if a translation fails to load it will notify the user. However, this code still fixes the original bug whereby scrolling in offline mode causes an unnecessary pop-up alerting the user they're offline. I avoid this bug simply by checking if the user is online/offline when loading translations.
> > > 
> > > 
> > > @stevekaplan123 so now if an offline user will tap translation and it will fail, he will have no message?
> > 
> > 
> > Probably yes, but that's a rare case, isn't it?
> 
> maybe, but this is not an edge case - he's using translations in the way he should and we have this error exactly for such cases

What you're suggesting does seem like an edge case.  I can't come up with a normal scenario for how this would happen.  If they're offline and don't have the Talmud downloaded, they will see the error message when they go to the Talmud and will never have a chance to try to download translations they don't have.  The only scenario would be if their wifi suddenly went out when they tried to load a translation, but if it stays off, they will get a message eventually from the main text which they won't be able to scroll up or down.

YishaiGlasner:
> > > > > @YishaiGlasner I removed the "fail silently" code so that now if a translation fails to load it will notify the user. However, this code still fixes the original bug whereby scrolling in offline mode causes an unnecessary pop-up alerting the user they're offline. I avoid this bug simply by checking if the user is online/offline when loading translations.
> > > > 
> > > > 
> > > > @stevekaplan123 so now if an offline user will tap translation and it will fail, he will have no message?
> > > 
> > > 
> > > Probably yes, but that's a rare case, isn't it?
> > 
> > 
> > maybe, but this is not an edge case - he's using translations in the way he should and we have this error exactly for such cases
> 
> What you're suggesting does seem like an edge case. I can't come up with a normal scenario for how this would happen. If they're offline and don't have the Talmud downloaded, they will see the error message when they go to the Talmud and will never have a chance to try to download translations they don't have. The only scenario would be if their wifi suddenly went out when they tried to load a translation, but if it stays off, they will get a message eventually from the main text which they won't be able to scroll up or down.

honestly, it's a product decision. we can tell the user the failue when he's tapping the translations, but the question is if we want to make the effort. so i'm approving this, for there is no code issues.

YishaiGlasner (dismissed):

YishaiGlasner:
> > > > > > @YishaiGlasner I removed the "fail silently" code so that now if a translation fails to load it will notify the user. However, this code still fixes the original bug whereby scrolling in offline mode causes an unnecessary pop-up alerting the user they're offline. I avoid this bug simply by checking if the user is online/offline when loading translations.
> > > > > 
> > > > > 
> > > > > @stevekaplan123 so now if an offline user will tap translation and it will fail, he will have no message?
> > > > 
> > > > 
> > > > Probably yes, but that's a rare case, isn't it?
> > > 
> > > 
> > > maybe, but this is not an edge case - he's using translations in the way he should and we have this error exactly for such cases
> > 
> > 
> > What you're suggesting does seem like an edge case. I can't come up with a normal scenario for how this would happen. If they're offline and don't have the Talmud downloaded, they will see the error message when they go to the Talmud and will never have a chance to try to download translations they don't have. The only scenario would be if their wifi suddenly went out when they tried to load a translation, but if it stays off, they will get a message eventually from the main text which they won't be able to scroll up or down.
> 
> honestly, it's a product decision. we can tell the user the failue when he's tapping the translations, but the question is if we want to make the effort. so i'm approving this, for there is no code issues.

maybe the fetching func should not pop the errro, but pass it to the caller, and the caller should decide about popping message. in this way we have full flexibility

PR #215, opened 2026-07-13
Fix bug when loading Sefaria on a new iOS device

nsantacruz:
This pull request introduces improvements to error logging and bug fix:

**Error Logging and Monitoring Improvements:**

* Enhanced the `ErrorBoundaryFallbackComponent` to log errors to Crashlytics, including details about downloaded books, which will help with debugging and monitoring in production. The component now also gracefully handles failures in retrieving the local book list and reports secondary errors to development logs. (`ErrorBoundaryFallbackComponent.js`)

**Bug Fix:**

* Modified the `FileSystem.makeDirectoryAsync` call in the `History` migration function to use the `intermediates: true` option, preventing errors if parent directories do not exist. (`history.js`). This was a very difficult bug to find that occurred when users bought a new iOS device and then downloaded Sefaria (which would retrieve the backup from iCloud). The app would error on startup due to trying to create the history directory.


ErrorBoundaryFallbackComponent.js
```diff
@@ -1 +1,30 @@
...
+
+export const ErrorBoundaryFallbackComponent = ({ error }) => {
+  useEffect(() => {
+    if (!(error instanceof Error)) { return; }
+
+    const logErrorToCrashlytics = async () => {
+      let downloadedBooks = [];
+      try {
+        downloadedBooks = await getLocalBookList();
+      } catch (e) {
```
YishaiGlasner: e is redundant and seems to be less recommended

ErrorBoundaryFallbackComponent.js
```diff
@@ -1 +1,30 @@
...
+        // best effort only, don't block error reporting on download-state lookup failing
+      }
+
+      await CrashlyticsService.recordError(error, {
```
YishaiGlasner: wrapping this with try and move the catch from l26 to here, will make it cleaner

ErrorBoundaryFallbackComponent.js
```diff
@@ -1 +1,30 @@
-export const ErrorBoundaryFallbackComponent = () => null;
\ No newline at end of file
+import { useEffect } from 'react';
```
YishaiGlasner, line 1: it seems that now it will be more appropriate to have it as jsx rather than js

YishaiGlasner (changes requested):

ErrorBoundaryFallbackComponent.js
nsantacruz: @YishaiGlasner I'm not sure why, but we don't use jsx in React Native.

ErrorBoundaryFallbackComponent.js
nsantacruz: Good catch, removed the unused binding — updated in c6b5a8cd.

ErrorBoundaryFallbackComponent.js
nsantacruz: Done — moved the catch here in c6b5a8cd.

ErrorBoundaryFallbackComponent.js
YishaiGlasner: ah, i forgot this. i thought this was js becaue it was kind of empty

YishaiGlasner (approved):

PR #216, opened 2026-07-19
docs: refresh dev-environment setup (README + setup skill)

yodem:
## What & why

Refreshes the developer setup docs to match the current toolchain (**RN 0.81.6 / React 19 / Expo modules 54**) and captures lessons from a full local setup that was **verified end-to-end** on macOS Apple Silicon — the app builds and launches on both the **iOS Simulator** and the **Android emulator** from a clean checkout.

Supersedes the stalled draft #147 (incorporates its Node/NVM, Xcode, and JDK notes).

## Changes
- **README.md**
  - Fix clone URL (`Sefaria-Mobile`, not `Sefaria-iOS`); drop the deprecated `npm run setup` (global `react-native-cli`) — RN 0.81 uses `npx react-native`.
  - Add prerequisites the old README omitted: **Ruby ~3.2 via rbenv** (system 2.6 fails `bundle install`), **JDK 17–20**, full **Xcode** + `xcodebuild -license accept`, and the required private **Firebase files** (with the correct bundle ids).
  - Document the Apple-Silicon iOS gotcha: `EXCLUDED_ARCHS[sdk=iphonesimulator*] = arm64` forces an x86_64/Rosetta build locally → build native arm64; and `pod deintegrate && pod install` when the Pods install is incomplete (GoogleUtilities header error).
  - Add a headless **Android SDK** install path (`sdkmanager`) as an alternative to the Studio wizard.
- **`.claude/skills/setup-mobile-dev/`** — a Claude Code skill that audits, installs, and repairs the dev environment (audit → auto-install → manual steps).
- **`.ruby-version`** — pins Ruby 3.2.2 for rbenv.

## Notes
- No app/source code changes — docs + tooling only.
- The `EXCLUDED_ARCHS = arm64` workaround in `ios/Podfile` is left as-is (CI runs on Apple Silicon, but the workaround is harmless there since CI only builds signed device archives, which don't use the `iphonesimulator*` SDK); the README just documents the local-arm64 override. Removing it repo-wide could be a follow-up.

https://claude.ai/code/session_014f2Sfjrd21kuxYss399tRm


dcschreiber (commented):
_(Claude writing on Daniel's behalf)_

Nice refresh — I fact-checked the doc claims against `master` (Gemfile Ruby/CocoaPods pins, `android/build.gradle` SDK/build-tools/NDK versions, both bundle ids, `ios/Podfile` `EXCLUDED_ARCHS`, the `index.js` whyDidYouRender gate, npm scripts) and they all hold up. The new `.ruby-version` is also safe for CI, since all three workflows pin `ruby-version: '3.2'` explicitly via `ruby/setup-ruby`. Three findings worth fixing before merge, each a one-line change:

**1. The "fine on Sefaria's Intel CI" claim is factually wrong (README note, PR body, and SKILL.md).**
All three workflows run on `macos-26` runners, which are Apple Silicon — GitHub retired Intel macOS runners. The workaround is harmless in CI for a different reason: CI only builds signed device archives (fastlane `build_app` + distribution cert), and the exclusion applies only to `iphonesimulator*` SDKs. Worth rewording, because the current text implies removing the workaround would break CI — it wouldn't, which actually strengthens this PR's own suggested follow-up of deleting the workaround repo-wide.

**2. `npm run setup` still exists in `package.json`.**
`"setup": "sudo npm install -g react-native-cli"` — the docs now correctly say it's deprecated, but the trap stays live for anyone who runs it out of habit, and it's a `sudo` global install of a package RN itself warns against. Deleting that one script line fits this PR's scope and makes the docs true by construction.

**3. Node floor is understated.**
README says "≥ 18 (20 recommended)", matching the repo's `engines: >=18` — but React Native 0.81 requires Node ≥ 20.19.4, and CI pins exactly `20.19.4`. A fresh dev on Node 18 will hit failures the docs say shouldn't happen. Suggest stating ≥ 20.19.4 (and ideally bumping `engines` to match, though that could be a follow-up).

---

**Nits** (take or leave, none blocking):

- README step 3 appends `rbenv init` to `~/.zshrc` unconditionally; the skill version has a `grep -q` guard — copying the guard into the README avoids duplicate lines on re-run.
- `sdkmanager --licenses` is interactive; the "headless" path should use `yes | sdkmanager --licenses` or it stalls in a script.
- SKILL.md points to the Sefaria wiki (`runbooks/sefaria-mobile-local-setup.md`) — this repo is public, so external contributors get a dead pointer to a private wiki. A parenthetical "(internal)" would help.
- The skill's disk-cleanup advice includes `pnpm store prune`, but this project uses npm.
- Version-rot risk is contained but real: "~121 pods", "iPhone 16", `Pixel_7_API_36` will drift. The skill already tells the agent to match `android/build.gradle`, which is the right instinct.

Overall: structure is a clear improvement (Requirements table + numbered steps), the skill's audit-first method is sound, no secrets exposed, `sudo` confined to genuinely manual steps. With 1–3 addressed this is good to merge.

dcschreiber (approved):

PR #217, opened 2026-07-21
feat: SSO sign-in, in-app forgot password, and auth hardening

yodem:
# SSO sign-in + auth hardening

## Blocking dependency
- ⚠️ **Sefaria-Project#3614 must merge first.** Two things on this branch don't work without it:
  - The in-app forgot-password request step is currently **CSRF-blocked** — it needs #3614's mobile-facing fix to submit at all.
  - #3614 also ships the `sso_only_account` contract on `api/login/` that this branch's email-login error handling consumes (see below).

## What this delivers

**Sign-in**
- Google & Apple SSO for register + login, native SDKs on iOS/Android where available (Apple on Android falls back to a web redirect)
- `auth.js` extracted out of `api.js` as its own module — all social/email login, JWT keychain storage, and token refresh now live there

**Error handling**
- Email login after an SSO-created account no longer fails silently: `login()` now reads the `sso_only_account` code off `api/login/`'s response and tells the user which provider(s) their account is linked to
- Raw SDK/error codes are no longer shown to the user — Apple's `1000` and friends now map to a plain network-error message (the code still reaches analytics and `__DEV__`)
- `password2` / `first_name` / `last_name` field errors from the register endpoint are surfaced individually instead of being swallowed into one generic banner

**Forgot password**
- New in-app "forgot password" screen (request step only) — previously this bounced to mobile web with no context for SSO users
- The emailed reset link still opens mobile web; routing that into the app would mean claiming `/password/reset*` as a universal link (out of scope here)

**Deep links**
- `login` / `register` / `logout` paths are now excluded from the app's deep-link handling, sent straight to the browser instead — the mobile counterpart to the AASA exclusion added in Sefaria-Project#3609, since those are browser-session/CSRF-bound flows the app can't complete

**Analytics**
- All auth events collapsed from the `sign_up_*` / `login_*` naming into one flat `auth_*` funnel (`auth_flow_started`, `auth_method_chosen`, `auth_process_started`, `auth_process_ended`, `auth_flow_ended`), matching the cross-platform spec so mobile and web report into the same funnel; a `flow_intent` field replaces what the old name prefix distinguished

**Reliability fixes found along the way**
- Offline cold start no longer signs the user out — a failed network refresh used to clear `isLoggedIn` even though `Sefaria._auth` stayed populated, leaving state and API calls disagreeing on login status. Only a server that actually rejects the refresh token signs the user out now.
- Legacy AsyncStorage → Keychain token migration no longer overwrites a newer sign-in's credentials
- SSO buttons now pass their provider into `onProcessEnded`, so switching focus to the email field mid-SSO-attempt no longer misattributes that event to the wrong flow

## Also in this branch
- Design tokens folded into `Styles.js`/theme files with named colors (was a separate `SSODesignTokens.js`)
- Apple Silicon simulator fix: dropped the stale `EXCLUDED_ARCHS[sdk=iphonesimulator*]` Intel workaround that forced simulator builds through Rosetta and blanked the Apple sign-in sheet
- Test harness repair — `master` currently has all 11 suites failing to load; this branch fixes that and adds coverage for the SSO handlers, cancellation paths, deep-link exclusions, password reset, and the auth event map

## Config / deploy notes
- iOS Google sign-in: resolved (OAuth client + `GOOGLE_SSO_IOS_CLIENT_ID` wired into all four deploy workflows)
- Apple sign-in against the backend needs Sefaria-Project#3590 merged + cauldron redeploy — no further mobile change required
- QA builds boot against production (no `BASE_HOST`), so testers repoint to `www.soo.cauldron.sefaria.org` (the `www.` is required) via the version-number tap-7-times menu

## Specs and decisions
- `docs/superpowers/specs/2026-07-21-mobile-sso-design.md`
- `docs/superpowers/specs/2026-08-04-sso-visual-qa-design.md`
- `docs/superpowers/specs/2026-08-13-in-app-password-reset-design.md`
- Full decision register: `wiki/meta/decision-2026-08-04-sso-mobile-open-decisions.md` (Sefaria wiki)

## Test plan
- [x] Buttons render on both platforms; Android sign-up shows Google only
- [x] RTL layout and Hebrew copy
- [x] Google sign-in end to end on Android and iOS against the soo cauldron
- [x] Analytics event sequence verified on device
- [ ] Apple sign-in on a physical iOS device — blocked on Sefaria-Project#3590
- [ ] Apple-on-Android web redirect round trip
- [ ] In-app forgot-password request — blocked on Sefaria-Project#3614 (CSRF)


analytics/authEventFamilies.js
```diff
@@ -0,0 +1,37 @@
...
+
+// Firebase Analytics caps event param VALUES at 100 chars.
+const FIREBASE_PARAM_VALUE_MAX_LENGTH = 100;
+
+// Truncates a value to Firebase's 100-char param value limit. Returns
+// undefined for nullish input so callers can omit the field.
+export const truncateForAnalytics = (value) => {
+  if (value === undefined || value === null) { return undefined; }
+  const str = String(value);
+  return str.length > FIREBASE_PARAM_VALUE_MAX_LENGTH ? str.slice(0, FIREBASE_PARAM_VALUE_MAX_LENGTH) : str;
```
YishaiGlasner:
```suggestion
  return str.slice(0, FIREBASE_PARAM_VALUE_MAX_LENGTH);
```

SSODesignTokens.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
```
YishaiGlasner: i'm not sure this file is needed. it's 19 consts, some of them are used once and some twice, so i suggest to have the consts in Styles.js and remove this file

ios/ReaderApp.xcodeproj/project.pbxproj
```diff
@@ -52257,7 +52265,6 @@
 				CURRENT_PROJECT_VERSION = 1;
 				DEAD_CODE_STRIPPING = YES;
 				DEVELOPMENT_TEAM = 2626EW4BML;
-				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = arm64;
```
YishaiGlasner, line 52260: is this safe? there was probably a reason this was excluded (this is also for l52360 removal).
update: it seems to break my local. Claude says that this is about Ruby:
gating on RUBY_PLATFORM at pod install time is fragile — it makes the resulting build architecture depend on which shell/Ruby happened to run pod install, not on what Xcode is actually targeting. Any dev or CI runner using an x86_64 Ruby on Apple Silicon (easy to hit by accident) reproduces exactly the mismatch you just got. A safer approach would be to drop the conditional entirely now that the stated purpose (support Apple Silicon simulators) is universal, or key it off something Xcode-controlled (like the EFFECTIVE_ARCH/NATIVE_ARCH build vars in a real build-setting override) rather than the installer's own host Ruby.

ThemeBlack.js
```diff
@@ -240,4 +240,34 @@ export default StyleSheet.create({
...
+    color: Colors.sefariaBlue,
+  },
+  ssoDividerLine: {
+    backgroundColor: Colors.secondaryBorder,
+  },
+  ssoDividerText: {
+    color: Colors.tertiaryText,
+  },
+  ssoErrorBannerBackground: {
+    backgroundColor: '#4A1F1F',
```
YishaiGlasner: use existing or new consts from Colors (and if there is incosistency flag it to product)

api.js
```diff
@@ -8,8 +8,37 @@ import 'abortcontroller-polyfill';
 import strings from './LocalizedStrings';
 import LinkContent from './LinkContent';
 import AsyncStorage from '@react-native-async-storage/async-storage';
+import * as Keychain from 'react-native-keychain';
 import { getCrashlytics, recordError } from '@react-native-firebase/crashlytics';  // to setup up generic crashlytics reports
 import jwt_decode from 'jwt-decode';
+import { devLog } from './devUtils';
+import { SSO_PROVIDER, AUTH_MODE, ANALYTICS_REASON, SSO_ERROR_CODE } from './AuthConstants';
```
YishaiGlasner: consider open a new auth.js file for all auth stuff and import it here

api.js
```diff
@@ -695,6 +724,117 @@ var Api = {
...
+        SSO_ERROR_CODE.NETWORK_ERROR,
+        ANALYTICS_REASON.NETWORK_ERROR,
+        { non_field_errors: `Network error during sign-in: ${error?.message}` },
+      );
+    }
+
+    // A redirect silently downgrades this POST to a GET, which hits a
+    // POST-only view and comes back with a non-JSON error body. Name it
+    // explicitly rather than letting it surface as a mystery 405.
+    if (response.url && response.url !== url) {
```
YishaiGlasner:
```suggestion
    if (response.redirected) {
```

api.js
```diff
@@ -733,21 +877,73 @@ var Api = {
...
+    return !!Sefaria._auth.uid;
   },
 
   getAuthToken: async function() {
     if (!Object.keys(Sefaria._auth).length) { return; /* logged out */ }
     const currTime = Sefaria.util.epoch_time();
     if (!Sefaria._auth.token || Sefaria._auth.expires <= currTime) {
-      const tempAuth = await AsyncStorage.getItem("auth");
-      Sefaria._auth = JSON.parse(tempAuth) || {};
+      await Sefaria.api._migrateLegacyAuthToken();
```
YishaiGlasner: Claude:
hydrateAuthFromKeychain() already runs the migration once at app init (StateManager.js). But getAuthToken() re-runs it again on every single expired-token refresh cycle for the entire life of the app session — an unconditional AsyncStorage.getItem call that will return null forever after the one relevant migration happened (or for any user who never had a legacy token). Harmless functionally, but it's dead weight in a hot path. Worth either dropping it from getAuthToken (rely on the one call in hydrateAuthFromKeychain) or gating it behind an in-memory "already migrated" flag.

SSOButtons.js
```diff
@@ -0,0 +1,257 @@
...
+        setLoadingProvider={setLoadingProvider}
+        onSSOSuccess={onSSOSuccess}
+        onSSOError={onSSOError}
+        onMethodChosen={onMethodChosen}
+        onProcessStarted={onProcessStarted}
+        onProcessEnded={onProcessEnded}
+        isHeb={isHeb}
+        theme={theme}
+      />
+      {showApple ? (
```
YishaiGlasner:
```suggestion
      {showApple && (
```
without `: null`

AuthPage.js
```diff
@@ -25,16 +23,57 @@ import Sefaria from './sefaria';
...
   setIsLoading(true);
   const mobileAppKey = await getMobileAppKey();
   formState.mobile_app_key = mobileAppKey;
   let errors = await Sefaria.api.authenticate(formState, authMode);
   if (!errors) { errors = {}; }
   setErrors(errors);
   setIsLoading(false);
-  if (Object.keys(errors).length === 0 && Sefaria._auth.uid) {
+  const success = Object.keys(errors).length === 0 && !!Sefaria._auth.uid;
+  onEmailSubmitResult?.(success);
```
YishaiGlasner: for what case we have the guarding quesion mark?

AuthPage.js
```diff
@@ -99,12 +220,63 @@ const AuthPage = ({ authMode, close, showToast, openLogin, openRegister, openUri
...
   const theme = getTheme(themeStr);
-  const isLogin = authMode === 'login';
+  const isLogin = authMode === AUTH_MODE.LOGIN;
   const placeholderTextColor = themeStr == "black" ? "#BBB" : "#777";
   const isHeb = interfaceLanguage === 'hebrew';
 
+  const [ssoError, setSsoError] = useState(null);
+  const emailCollisionMessage = ssoCollisionMessage(errors.email);
+
+  const handleSSOSuccess = async (provider, idToken, userData) => {
```
YishaiGlasner: this name is ok in `SSOButtons`, but here there is no meaning to the success (the reader thinks the sso had success, but hen the sso is made on the first line).
handleSSOTokenReceived would be more accurate

AuthPage.js
```diff
@@ -99,12 +220,63 @@ const AuthPage = ({ authMode, close, showToast, openLogin, openRegister, openUri
...
+      }
+    }
+  };
+
+  const handleSSOError = (error) => {
+    console.log('SSO Error:', error);
+    if (__DEV__) {
+      setSsoError(`SSO [${error?.code}] ${error?.message}`);
+    } else {
+      // The SDK's error message can be long and device-specific, so only the
```
YishaiGlasner: move this comment to l263

ReaderApp.js
```diff
@@ -1089,8 +1094,9 @@ class ReaderApp extends React.PureComponent {
       }
       this.modifyHistory({ dir: "forward", state: this.state });
     }
-    this.setState({menuOpen: menu});
-    if (via && typeof via === 'string') {
+    const source = (via && typeof via === 'string') ? via : null;
```
YishaiGlasner:
```suggestion
    const source = (typeof via === 'string') ? via : null;
```
typeof doesn't raise name error

SSOButtons.js
```diff
@@ -0,0 +1,257 @@
...
+  );
+};
+
+const SSOButtons = ({ authMode, onSSOSuccess, onSSOError, onMethodChosen, onProcessStarted, onProcessEnded }) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [loadingProvider, setLoadingProvider] = useState(null);
+  const { interfaceLanguage, themeStr } = useContext(GlobalStateContext);
+  const isHeb = interfaceLanguage === 'hebrew';
+  const theme = getTheme(themeStr);
+  const showApple = Platform.OS === 'ios' || authMode === AUTH_MODE.LOGIN;
```
YishaiGlasner: comment the product decision

ReaderApp.js
```diff
@@ -34,6 +34,7 @@ import styles from './Styles';
 import strings from './LocalizedStrings';
 import Sefaria from './sefaria';
 import { LinkFilter } from './Filter';
+import { AUTH_MODE } from './AuthConstants';
```
YishaiGlasner, line 37: use it also in `openMenu` (l1087)

analytics/__tests__/authEventFamilies.test.js
```diff
@@ -0,0 +1,47 @@
...
+import { AUTH_EVENT_FAMILY, truncateForAnalytics, generateUUID } from '../authEventFamilies';
+
+describe('AUTH_EVENT_FAMILY', () => {
+  test('maps register to sign_up', () => {
+    expect(AUTH_EVENT_FAMILY.register).toBe('sign_up');
+  });
+  test('maps login to login', () => {
+    expect(AUTH_EVENT_FAMILY.login).toBe('login');
+  });
+});
```
YishaiGlasner: this is almost meaningless

__tests__/AuthPage.test.js
```diff
@@ -48,14 +64,40 @@ describe('login', () => {
...
+      .toBe(strings.ssoEmailExistsGoogle);
+  });
+  test('returns null for an unrelated field error', () => {
+    expect(ssoCollisionMessage('This password is too short.')).toBeNull();
+  });
+  test('returns null for undefined/empty input', () => {
+    expect(ssoCollisionMessage(undefined)).toBeNull();
+    expect(ssoCollisionMessage(null)).toBeNull();
+  });
+});
```
YishaiGlasner, line 314: no tests for handleSSOSuccess and handleSSOError

__tests__/ssoGoogleCancel.test.js
```diff
@@ -0,0 +1,81 @@
+// Generated by OrchestKit Claude Plugin
```
YishaiGlasner: we should have parallel tests for apple

YishaiGlasner (changes requested):

ios/ReaderApp.xcodeproj/project.pbxproj
yodem: Agreed on the diagnosis, and that's actually what's already here — the `RUBY_PLATFORM` gate was an intermediate step (06c29067) and it's gone. `640eea4e` took your second suggestion and dropped the conditional entirely, on both sides:

- `ios/Podfile` — the whole `installer.pods_project.build_configurations.each` block that set `EXCLUDED_ARCHS[sdk=iphonesimulator*]` on the **Pods** project
- `ios/ReaderApp.xcodeproj/project.pbxproj` — the same key on both app-target configs (l52260 and l52360)

Dropping both together is the point: the earlier host-Ruby gate could leave the Pods project and the app target disagreeing on a real Intel host, which is the exact mismatch you're describing. Stock RN 0.81 sets no `EXCLUDED_ARCHS` at all; the setting dates to c6af30a5 ("bigsur edits", Feb 2021) as an Intel-Mac workaround.

Why it had to go rather than stay: on Apple Silicon it forces the simulator slice to x86_64, so the app runs under Rosetta — and `ASWebAuthenticationSession` renders a **blank sheet** under Rosetta. That's the white screen after tapping "Continue with Google". It also makes arm64-only runtimes (iOS 26+) untargetable. Simulator-only either way; device and TestFlight builds were never affected.

On your local break — that reads like stale artifacts rather than the setting itself, since the old x86_64 Pods are still on disk and won't relink against an arm64 slice. Could you try:

```
cd ios && rm -rf Pods Podfile.lock ~/Library/Developer/Xcode/DerivedData/ReaderApp-* && pod install
```

then a clean build. If it still breaks after that, can you paste the actual error and tell me whether you're on Apple Silicon or Intel? If it's genuinely Intel-specific I'd rather fix it properly with an Xcode-controlled build setting, as you suggested, than restore the host-Ruby gate.

analytics/authEventFamilies.js
yodem: Applied. `slice` already returns the whole string when the end index is past the end, so the ternary was doing nothing. Reworded the comment above it to say that, so it does not get "defensively" restored later.

SSODesignTokens.js
yodem: Done — file deleted, the consts moved into `Styles.js` next to the existing top-of-file consts (`readerSideMargin` et al), and the `export` keyword dropped since nothing outside consumed them.

I kept the comments that carry provenance rather than the values: the Figma spec path, the per-token Figma variable names (`--sds-size-space-600` and friends), and the note that there is deliberately no fixed container width. That last one is load-bearing — a fixed 337pt width copied off the Figma frame is exactly the bug `__tests__/ssoLayout.test.js` exists to prevent.

ThemeBlack.js
yodem: Applied — no bare hex left in the SSO block of either theme. Two mapped onto existing consts exactly (`ssoDividerLine` → `Colors.lightGrey` `#ccc`, `ssoDividerText` → `Colors.secondaryText` `#999`); the rest got new named consts rather than being snapped to a near-match.

Flagging four things for product/design, per your parenthesis:

1. **The SSO button stays `#FFFFFF` in dark mode too.** Google brand guidelines want the colour mark on white, and `img/sso-apple.png` is solid black with no light-on-dark variant — so it renders as a white island in dark mode. Named `ssoButtonSurface` rather than pointed at a palette colour, because it has to stay white in both themes. Product call: ship a dark-mode Apple mark and accept a dark button, or keep it.
2. **The Figma spec has no dark-mode divider values** at all — it gives `#CCCCCC` / `#999999` for light only. Dark currently uses `secondaryBorder` `#666` / `tertiaryText` `#bbb`, which is at least palette-consistent, but design has never actually specified it.
3. **The error banner colours are in no spec.** Six values across the two themes, none of which appear in the Figma colour table — and the palette had **no** error/red constant of any kind before this branch. They are now `errorBannerBackground` / `errorBannerBorder` / `errorBannerText`, which means they will quietly become the app-wide error colours by default. Worth design ratifying rather than inheriting.
4. Minor, non-colour: `ssoButtonPressed: { opacity: 0.2 }` — the wireframe defines no pressed state, so it borrows `TouchableOpacity`'s default. Comment already notes this in `Styles.js`.

api.js
yodem: Done — new `auth.js`, and `api.js` drops from 1033 to 695 lines.

Moved: the keychain constants, `SSO_PROVIDER_CONFIG`, `login` / `register` / `socialLogin` / `socialLoginRequest` / `refreshToken` / `authenticate`, the token storage layer (`storeAuthToken`, `getAuthToken`, `hydrateAuthFromKeychain`, `clearAuthStorage`, `_migrateLegacyAuthToken`) and `deleteUserAccount`.

Pure move — no behaviour change beyond the two fixes in your other comments. `api.js` spreads `Auth` into the `Api` object, so every `Sefaria.api.<method>` name is unchanged and no call site or test import moved.

Two deliberate choices worth a second opinion:
- **No cycle.** `auth.js` does not import `api.js`; it reads `Sefaria.api._baseHost` off the global at call time, as before. Inter-method calls still go through `Sefaria.api.*` rather than direct local references — that is what keeps the existing test stubs working, so I commented it to stop it being tidied away.
- **`deleteUserAccount` came along.** It is arguably account management rather than auth, but it fetches the token and builds an authorised request, and it lived in the auth block. Happy to move it back if you would rather.

api.js
yodem: Applied, with the gating flag rather than the removal — dropping the call outright would have broken something non-obvious.

The migration deletes the legacy AsyncStorage copy **only after** the keychain write succeeds. On a locked device that write throws, and the retry-on-next-read through `getAuthToken` is the entire recovery path; without it that user is logged out with no way back. So `getAuthToken` has to keep calling it.

The flag therefore means *"the legacy key is confirmed gone"*, not *"we tried once"*. Set on four definitive outcomes: no legacy value; malformed JSON; parsed but no `.token`; or a successful keychain write followed by the delete. Explicitly **not** set when the keychain write throws — that path returns early with the legacy copy still on disk and retries next cycle.

Net result is what you were after: after the first call the hot path skips the `AsyncStorage.getItem` entirely, for everyone except the user who actually needs the retry. Comments at the declaration, at the failed-write `return`, and at the call site, since this is exactly the subtlety someone will optimise away later.

SSOButtons.js
yodem: Applied. Safe here since `showApple` is `Platform.OS === 'ios'`, a strict boolean — no risk of leaking a stray `0` or `''` into the tree.

AuthPage.js
yodem: No case — removed. I traced it: `onEmailSubmitResult` is a parameter of the module-private `onSubmit`, whose only caller is the closure returned by `useAuthForm`, which always passes all six arguments. `useAuthForm` is also module-private with exactly one call site, and that one always passes `handleEmailSubmitResult` — an unconditionally-defined `const`. Neither function is exported. So it could never be undefined, and the `?.` was hiding nothing.

AuthPage.js
yodem: Renamed to `handleSSOTokenReceived` — you are right, the first line of it is the backend exchange, which is the thing most likely to fail. Kept the `onSSOSuccess` prop name on the `SSOButtons` side as you suggested, so it now reads `onSSOSuccess={handleSSOTokenReceived}` — the provider succeeded, what happens next is ours.

Left `handleSSOError` alone: it receives an actual error and only renders it, so that name does match what it does.

AuthPage.js
yodem: Moved. Line 263 at the time you commented was the first `ssoErrorWithCode` call site — the one in the success handler — so the "why only the code, never the raw message" rationale now sits where the decision first appears, and points forward at the second call site in `handleSSOError`.

Reworded slightly in the process: at the new position the code is a **server** error code rather than an SDK one, so it covers both sources now instead of describing only the SDK case.

SSOButtons.js
yodem: Already done — this thread went stale against `93ee5028`, which landed the rationale as a comment block above `showApple`:

```js
  // TEMPORARY: Apple sign-in is hidden on Android, iOS is unaffected.
  //
  // On Android there is no native Apple SDK, so the button hands off to mobile
  // web (see AppleSignInButton's else branch). That round trip signs the user in
  // on the web session only -- there is currently no way to bring them back into
  // the app authenticated, so an Android user who originally registered with
  // Apple has no working way in and lands in a dead end. Hiding the entry point
  // is better than offering a path that cannot complete.
  //
  // Deliberately hidden, not deleted: the Android redirect path below stays in
  // place so this becomes a one-line revert once the deep-link-back is built.
  // Sizing that work is queued behind the web SSO release.
```

The blocker is backend/deep-link work rather than mobile UI, so restoring the button is a one-line revert once that exists. Full decision register, including the go/no-go sequencing, is in the wiki at `wiki/repos/sefaria-mobile/sso.md`.

api.js
yodem: Pushing back on this one — `response.redirected` does not exist on React Native, so the guard would silently never fire.

RN's `fetch` is the whatwg-fetch polyfill re-exported as-is:
- `node_modules/react-native/Libraries/Network/fetch.js` — `require('whatwg-fetch')`, then `export const Response = global.Response;`. No wrapper, no added fields.
- `node_modules/whatwg-fetch/package.json` — version **3.6.20**.
- `grep -n redirected node_modules/whatwg-fetch/fetch.js` — **zero matches** in the whole file. The `Response` constructor sets exactly `type`, `status`, `ok`, `statusText`, `headers`, `url` and the body; `clone()` copies only `status` / `statusText` / `headers` / `url`.

So `if (response.redirected)` compiles fine, is `undefined` at runtime, and the redirect case falls through to be reported as an ordinary failure. That is the failure this check exists to catch: the cauldron 301-redirects a missing `www.`, and a redirect downgrades the SSO POST to a GET, so the endpoint fails in a way that looks like nothing in particular.

The polyfill's only redirect signal is the URL itself — line 552: `options.url = 'responseURL' in xhr ? xhr.responseURL : options.headers.get('X-Request-URL')`, i.e. XHR's post-redirect final URL lands in `response.url`. Hence the existing `response.url && response.url !== url` comparison, which is the only form that actually works here.

Left as-is, but I have added a comment at the check recording the polyfill version and this reasoning, so the next reader has the answer in place instead of re-deriving it. Happy to revisit if we ever move off the whatwg polyfill.

ReaderApp.js
yodem: Took the shape of this but kept one guard — landed on `const source = (typeof via === 'string' && via) ? via : null;`.

You are right that `typeof` is the safe test and that leading with `via &&` for that purpose was redundant. But the `&&` was also doing a second job: rejecting the empty string.

Checking every `openMenu(` call site, the reason the `typeof` check exists at all is that `openLogin` / `openRegister` get handed **straight to `onPress`** in three places — `AuthPage.js:369` and `:382`, `SwipeableCategoryList.js:188`, `HistorySavedPage.js:369` — so RN invokes them with a `PressEvent` object, which flows into `openMenu(menu, via)`. Without the guard that object would land in `menuOpenSource` and in the `trackEvent("OpenMenu", {menu, via})` payload.

Nothing passes `''` today, so your version would not break anything right now — the trailing `&& via` is just keeping a blank source out of the analytics event if someone later threads a value through. Say the word if you would rather have the plain `typeof` form and I will drop it.

Also picked up your other point on this function: `SKIP_MENUS` now uses `[AUTH_MODE.REGISTER, AUTH_MODE.LOGIN]` (verified byte-identical to the old literals) and moved to module scope, since it was being reallocated on every call.

ReaderApp.js
yodem: Done — `SKIP_MENUS` is now `[AUTH_MODE.REGISTER, AUTH_MODE.LOGIN]`. Confirmed the constants are byte-identical to the literals they replace (`AuthConstants.js`: `LOGIN: 'login'`, `REGISTER: 'register'`), so no behaviour change.

Also hoisted it to module scope while I was in there — it was a fixed array being reallocated on every `openMenu` call, and `ViewPort` is the existing precedent for a module-level const in this file.

Swept the rest of `ReaderApp.js` for other bare `"register"` / `"login"` menu literals: none left, lines 1222/1226 were already using `AUTH_MODE`. There is still one bare `'login'` in `HistorySavedPage.js:85`, which I left alone as out of scope for this PR — happy to pick it up if you want it.

analytics/__tests__/authEventFamilies.test.js
yodem: Agreed — it asserted a two-entry map had the two values written directly above it in the same file. Replaced with three invariants driven off `Object.values(AUTH_MODE)`:

1. every `AUTH_MODE` value maps to a non-empty string family — so adding a third mode without a family entry fails loudly instead of silently emitting `undefined_flow_started` at runtime;
2. the map's keys are exactly the `AUTH_MODE` values — no orphan entries left behind after a rename;
3. the families are distinct — the sign-up and login funnels stay separable in the analytics.

Those keep being meaningful when a mode is added, which the old block did not.

Also dropped the stray `// Generated by OrchestKit Claude Plugin` header from this file and from `ssoGoogleCancel` / `ssoAppleAndroid` while I was in there.

__tests__/AuthPage.test.js
yodem: Added — 7 tests in a new `describe('SSO handlers')`, driving both handlers through the rendered component's `onSSOSuccess` / `onSSOError` props and reading what the user actually sees off the rendered error banner, rather than calling the handlers directly.

- **`handleSSOTokenReceived` success** — `socialLogin` called with `(provider, idToken, userData)`, then `syncProfile` / `close(authMode)` / the success toast; no banner.
- **`handleSSOTokenReceived` failure** — the banner is exactly `ssoErrorWithCode(result.code)`, and the raw server message is asserted *absent*. Plus a clamping case: a server-controlled code containing `<script>` never reaches the banner.
- **`handleSSOError`** — banner comes from `error.code`, SDK message not surfaced; a code-less error falls back to `ssoErrorGeneric`.
- **Stale-error regressions** — the two bugs this PR fixed now have guards: after an SSO error, a register submit hitting the Google collision sentence shows `ssoEmailExistsGoogle` (the stale error no longer masks it), and a clean submit clears the banner.

One thing worth knowing for anyone extending these: Jest sets `__DEV__ = true`, which selects AuthPage's developer branch that deliberately prints raw server/SDK text. The suite pins `global.__DEV__ = false` per-test and restores it, so the assertions are about shipped behaviour, not dev behaviour. Without that pin these tests would have passed while asserting the wrong thing.

__tests__/ssoGoogleCancel.test.js
yodem: Added `__tests__/ssoAppleCancel.test.js`, 6 tests mirroring this file's structure.

The contracts differ enough that it is not a copy-paste: Google v13 *resolves* with `{type:'cancelled'}`, whereas `@invertase/react-native-apple-authentication` *rejects* with `Error.CANCELED` — verified as `'1001'` at `node_modules/@invertase/react-native-apple-authentication/lib/index.d.ts:140` rather than assumed. The mock also needs `__esModule: true` + `default`, since `SSOButtons` reaches for `.default`. And since Apple is now hidden on Android, the suite pins `Platform.OS = 'ios'` or there is no button to press.

Parallel to the Google cases: a cancel raises no banner; a cancel records `FAILURE` / `CANCELLED` rather than a provider error; a success forwards the identity token. Three Apple-specific ones on top:

- a **non-cancel** rejection is a real `PROVIDER_ERROR` and does call `onSSOError`;
- **repeat sign-in** — Apple only returns email/fullName on the very first authorization, so a response with `email: null, fullName: null` must still forward the token (this is the case that silently breaks if someone starts trusting `userData.email`);
- a response with no identity token is a `PROVIDER_ERROR`, not a silent no-op.

Left the `makeHandlers` / press-helper duplication between the two files rather than extracting a shared helper — there is no `__tests__/helpers` convention in this repo to slot it into. Say the word if you would rather have one.

ios/ReaderApp.xcodeproj/project.pbxproj
YishaiGlasner: it's working after pod install. sorry

ThemeBlack.js
YishaiGlasner: you flagged them to Penina, right?

__tests__/ssoAppleCancel.test.js
```diff
@@ -0,0 +1,136 @@
+// Generated by OrchestKit Claude Plugin
```
YishaiGlasner: remove

yodem:
### Added: forgot-password in-app screen + auth error fixes

Three commits on top of the SSO work.

**`1c1d56f9` — deep link exclusions.** `login`, `register` and `logout` matched the generic single-segment route and were treated as possible text titles, reaching the browser only after a failed `api/name` lookup. They now route straight out. This is the mobile counterpart to the AASA exclusion list in Sefaria-Project#3609 — the two lists must change together, and neither repo's tests can see the other.

**`bd6d0f3e` — auth errors.** Fixes three user-visible failures:
- Email login after SSO failed *silently*: `api/login/` is stock SimpleJWT and returns `{"detail": ...}`, which no render path covered. Unrecognised keys now fall back to a visible message, and the `sso_only_account` contract maps the account's providers to the right copy. (Depends on Sefaria-Project#3614 to actually name the provider on this path.)
- Apple's `1000` was surfacing raw as `יש תקלה, נסו שוב 1000`; it now shows the network string.
- `password2`, `first_name` and `last_name` had no entry in the known-field list, so the new fallback would have covered their real messages with a generic banner.

**`26f31871` — forgot-password screen.** Request step as a third `AuthPage` mode, per Penina's mock. The form persists; an SSO-only account surfaces in the existing error banner above it with a per-provider link that runs native sign-in and logs the user into the app. Success replaces the form with "Reset Link Sent". Copy is verbatim from web's interface strings plus the app-specific sheet.

---

### ⚠️ Blocked on a backend change — the reset request will 403 in production

`sso/views.py::password_reset_api` has `@require_POST` but **not** `@csrf_exempt`, unlike `google_mobile` and `apple_mobile` which both have it. `CsrfViewMiddleware` is enabled globally. The app is JWT-only and holds no `csrftoken` cookie, so every "Send Reset Link" tap returns 403 with an HTML body, which fails JSON parsing and surfaces as the generic error.

Found in review, not on device — the unit tests mock `fetch`, so nothing here could catch it. **The screen cannot work until `password_reset_api` is exempted** (or the app is given a CSRF token). Worth deciding which: exempting matches the existing mobile endpoints, but that endpoint is also called from web, where the cookie is present today.

### Scope note

Request step only. The emailed confirm link still opens mobile web — routing it into the app would mean claiming `/password/reset*` as a universal link, which is exactly the exclusion added to stop the app capturing auth URLs mid-flow.

### Open with the designer
- "Reset Link Sent" has no "Back to login" and no button — only the close control, so a user has to close and reopen the auth page to get back to login. Implemented as drawn; flagged to Penina.

ios/ReaderApp.xcodeproj/project.pbxproj
yodem: Good, thanks for confirming.

ThemeBlack.js
yodem: No — I do not have evidence these were flagged to Penina in this session. Still needs doing. Can you confirm, or should I open that with her?

__tests__/ssoAppleCancel.test.js
yodem: Confirmed — lines 1-2 are a leftover generator header comment ("// Generated by OrchestKit Claude Plugin" / "// Created: ..."), not needed. Same header pattern shows up at the top of auth.js too. This needs a code edit I cannot make in this session; flagging it back so the other agent handling edits removes it.

auth.js
```diff
@@ -0,0 +1,541 @@
...
+    const url = `${Sefaria.api._baseHost}api/register/`;
+    const authBody = {
+      email: authData.email,
+      first_name: authData.first_name,
+      last_name: authData.last_name,
+      password1: authData.password,
+      password2: authData.password,
+      mobile_app_key: authData.mobile_app_key,
+    };
+    console.log(authBody, Sefaria.api.urlFormEncode(authBody));
```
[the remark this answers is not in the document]
yodem: Confirmed — register() logs authBody, which includes password1/password2 in plaintext. Being fixed now (removing/redacting that console.log); not merged yet.

SSOButtons.js
```diff
@@ -0,0 +1,314 @@
...
+      // there is no signal back into the app tied to the actual sign-in
+      // outcome, so we deliberately fire method_chosen + process_started only
+      // and nothing further here. Analytics infers abandonment from a
+      // flow_started with no matching flow_ended.
+      onProcessStarted();
+      try {
+        await Linking.openURL(appleAndroidRedirectUrl());
+      } catch (error) {
+        onSSOError(new Error('Could not open Apple sign-in. Please try again.'));
+      }
```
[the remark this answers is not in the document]
yodem: Agreed — the Android Apple redirect catch calls onSSOError but never onProcessEnded, so a Linking.openURL failure leaves that attempt without a closing analytics event. Being fixed now to emit onProcessEnded(FAILURE) before onSSOError in that catch; not merged yet.

AuthPage.js
```diff
@@ -25,16 +23,191 @@ import Sefaria from './sefaria';
...
+);
+
+// Builds the ERROR view's banner content, highest precedence first: a live
+// ssoError from tapping one of this banner's own provider links; then an
+// sso_only_account result, one message+link row per linked provider (Apple
+// excluded on Android, same as SSOButtons' own showApple); else the generic
+// message with no link.
+const forgotPasswordBannerError = (auth, ssoError, { showAppleLink, disabled, onGoogleLink, onAppleLink }) => {
+  const rows = [];
+  if (auth && auth.code === AUTH_ERROR_CODE.SSO_ONLY_ACCOUNT) {
```
YishaiGlasner:
```suggestion
  if (auth?.code === AUTH_ERROR_CODE.SSO_ONLY_ACCOUNT) {
```

AuthPage.js
```diff
@@ -99,36 +391,200 @@ const AuthPage = ({ authMode, close, showToast, openLogin, openRegister, openUri
...
+      console.error('Unexpected error requesting password reset:', error);
+      setForgotPasswordSsoAuth(null);
+      setForgotPasswordView(FORGOT_PASSWORD_VIEW.ERROR);
+    } finally {
+      setForgotPasswordIsLoading(false);
+    }
+  };
+
+  const forgotPasswordBanner = forgotPasswordView === FORGOT_PASSWORD_VIEW.ERROR
+    ? forgotPasswordBannerError(forgotPasswordSsoAuth, ssoError, {
```
YishaiGlasner: use &&

auth.js
```diff
@@ -0,0 +1,612 @@
...
+      body: JSON.stringify(config.buildBody(idToken, userData)),
+    });
+  },
+  // `analyticsError` is the ANALYTICS_REASON fallback AuthPage reports as the
+  // analytics `error` field when `code` itself isn't a usable value (e.g. the
+  // SERVER_REJECTED path below, where `code` is `data.error` and may be absent).
+  _socialLoginFailure: function(code, analyticsError, error) {
+    return { success: false, code, analyticsError, error };
+  },
+  socialLogin: async function(provider, idToken, userData) {
```
YishaiGlasner: isn't this and requestPasswordReset can be one function?

auth.js
```diff
@@ -0,0 +1,612 @@
...
+    // its content instead of just a status code, same as socialLogin. A bare
+    // 200 {} success body has no JSON worth parsing.
+    //
+    // Guarded (unlike socialLogin's equivalent read) because a body-stream
+    // failure mid-read here would otherwise reject past requestPasswordReset's
+    // classify-don't-throw contract; classified the same way socialLogin
+    // classifies its own fetch() failure, for the same reason.
+    let rawBody;
+    try {
+      rawBody = await response.text();
```
YishaiGlasner: can this and (l314-319) be moved into the first try block, as the error catching is similar?

AuthPage.js
```diff
@@ -99,36 +391,200 @@ const AuthPage = ({ authMode, close, showToast, openLogin, openRegister, openUri
...
+  const handleSSOError = (error) => {
+    console.log('SSO Error:', error);
+    if (__DEV__) {
+      setSsoError(`SSO [${error?.code}] ${error?.message}`);
+    } else {
+      setSsoError(ssoErrorWithCode(error?.code));
+    }
+  };
+
+  // --- Forgot-password mode (AUTH_MODE.FORGOT_PASSWORD) ---------------------
```
YishaiGlasner: **Extract forgot-password into its own component + hook(s)**

`AuthPage.js` is 806 lines and now covers all three `AUTH_MODE`s in one component. Forgot-password's own state/handlers/JSX (lines ~476-715, ~240 lines / ~30% of the file) is large enough to pull into its own component — and since this PR is the one that added the third mode and already started this exact pattern with `useSSOSignIn`, it's the natural place to finish the split rather than let it accrete further.

The catch: forgot-password's SSO-only banner deliberately shares the same flow-level analytics wiring as login/register — `fireMethodChosen`, `fireProcessStarted`, `fireProcessEnded`, `handleSSOError`, `onSSOSignInSuccess` (all closures over `AuthPage`'s `flowIdRef`/`attemptIdsRef`/`flowOutcomeRef`), plus the shared `ssoError` state — so a provider tap from this screen reports into the same attempt funnel as the main screen. A clean extraction needs that bookkeeping lifted into its own hook (or folded into `useSSOSignIn`) rather than prop-drilled wholesale into a new `<ForgotPasswordScreen>`.

Suggest: a hook (e.g. `useAuthAnalytics` or similar) owning the flow/attempt refs and the four `fire*` functions, returned to both `AuthPage` and the new forgot-password component; the component itself then just needs `theme`, `isHeb`, `close`, `openLogin`, and the analytics hook's output.

YishaiGlasner (changes requested):

__tests__/ssoAppleCancel.test.js
yodem: Removed in `5cdc5ec4`. The `// Generated by OrchestKit Claude Plugin` banner was a plugin artifact, not something the test needed — no other file in the repo carries it now either.

auth.js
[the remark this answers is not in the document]
yodem: Fixed in `5cdc5ec4`. The `console.log(authBody, ...)` is gone — it was logging `password1`/`password2` in cleartext to logcat and Console.app on every registration. Not redacted, removed: there was nothing in it worth keeping, since the same call already reports failures through the error path.

The only `console.log` left in `auth.js` is the `"Not signed in"` guard, which carries no user data.

SSOButtons.js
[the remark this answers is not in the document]
yodem: Fixed in `5cdc5ec4` — the `catch` now bookends the attempt before surfacing the error:

```js
onProcessEnded({ status: ANALYTICS_STATUS.FAILURE, error: ANALYTICS_REASON.PROVIDER_ERROR }, SSO_PROVIDER.APPLE);
onSSOError(new Error("Could not open Apple sign-in. Please try again."));
```

To be explicit about the case you flagged as intentional: the *successful* redirect still fires `process_started` with no `process_ended`, and that is deliberate. Apple has no native Android SDK, so this path is a `Linking.openURL` out of the app and nothing signals the real outcome back. Abandonment is inferred from a `flow_started` with no matching `flow_ended`; a synthetic success there would be a lie. A failure to even open the browser is different — the app knows it happened, so it now closes the attempt.

ThemeBlack.js
yodem: Checked this properly and the answer turned out to be "there is nothing to flag" — with a more interesting finding underneath.

**On product:** every colour in the SSO block comes from the Figma design, including the error banner's fill and border. The light-only treatment is not an oversight either — it is a deliberate decision Penina and Yotam took: this one page is light-only on mobile, while other pages keep both themes. So the Figma file containing only a white theme is the spec, not a gap in it. She is aware; there is no open question for her.

I did compare against web while checking, for completeness: web renders auth errors as red text (`#c03522` / `$sef-error`) with an icon and no fill, whereas the app renders a filled box. That is a real difference, but it is a deliberate one she drew, not drift.

**The finding:** the block you commented on is unreachable. `AuthPage` pins itself to the light theme (`AUTH_PAGE_THEME = 'white'`) and passes that down as a prop to every child, so `ThemeBlack`'s `sso*` keys are never read — `ssoLayout.test.js` already asserts the light theme is used for *both* app theme settings. The dark values I had written there render nowhere and cannot.

I kept the block rather than deleting it, so `ThemeBlack` does not silently lack keys `ThemeWhite` has: if `AUTH_PAGE_THEME` ever goes away, the page degrades to mistuned colours instead of `undefined` styles. It now carries a comment saying it is unreachable, why, and that the colours need design sign-off before they could ever go live — otherwise the next person tunes them against a screen they cannot produce.

Happy to just delete it instead if you would rather not carry dead code.

AuthPage.js
yodem: Pushing back on this one — `&&` would introduce a PropTypes warning.

This is not a JSX render position; it assigns `forgotPasswordBanner`, which is then passed as the `error` prop to `SSOErrorBanner`. With a ternary the non-error case is `null`; with `&&` it becomes `false`.

`SSOErrorBanner.propTypes.error` is a `PropTypes.shape({...})`. `shape` short-circuits on `null`/`undefined` but validates anything else, so `false` fails the type check:

```
$ node -e "PropTypes.checkPropTypes({error: shape}, {error: false}, ...)"
Warning: Failed prop type: Invalid prop `error` of type `boolean` supplied to `SSOErrorBanner`, expected `object`.
```

`null` produces no warning. The runtime behaviour is unaffected either way — the component does `if (!error) return null` — so this is purely about not adding a dev-console warning on every non-error render of the forgot-password screen.

Left as the ternary. Happy to switch if you would rather have `&&` and relax the propType to `PropTypes.oneOfType([shape, PropTypes.bool])`, but that seems like the worse trade.

auth.js
yodem: Done in `9ef7d6cc` — not merged into one public function, but the duplication you spotted is gone.

They cannot become a single function: different signatures, different success paths (`socialLogin` stores tokens and handles `sso_only_account`; `requestPasswordReset` just reports success), different callers. What they *did* duplicate is the whole middle: fire the request, guard the POST→GET redirect, read the body as text, parse it as JSON, classify a failure at each stage.

That is now `_postAndReadJson`, which takes the caller's failure factory and its own per-stage wording and returns either `{failure}` or `{response, data}`. Each public function is left with only what genuinely differs. Net −121/+79 lines.

Deliberately kept distinct: the diagnostic strings. "Network error during sign-in" vs "Network error requesting password reset" land in Crashlytics, and collapsing them would make the two indistinguishable in a crash report.

auth.js
yodem: Folded, via the same helper — `9ef7d6cc`.

All three stages (request, redirect check, body read) now live in one linear sequence inside `_postAndReadJson` rather than three separate blocks in each caller, which is the consolidation you were after.

One deliberate deviation: I did not nest the redirect check inside the request's `try`. It is a plain `if`, not a throwing operation, so putting it there would imply it can throw without changing any behaviour. It sits immediately after the try instead.

Your comment also turned up a real bug while I was in here. Extracting the helper meant passing the request in, and the natural way — `_postAndReadJson(Sefaria.api.socialLoginRequest(...), ...)` — evaluates the fetch at the *call site*, outside any `try`. A rejection is still caught, but a **synchronous** throw would escape `socialLogin` entirely and break the classify-don't-throw contract. It now takes a thunk invoked inside the try. Added a test that fails against the promise form, since nothing in the suite covered a sync throw.

AuthPage.js
yodem: Applied in `48a4cfcf`. Optional chaining is already the prevailing style in this file, so `auth && auth.code` was the odd one out.

The function moved to `ForgotPasswordScreen.js` in the same commit (see the extraction thread), so it now reads `if (auth?.code === AUTH_ERROR_CODE.SSO_ONLY_ACCOUNT)` there.

AuthPage.js
yodem: Done in `48a4cfcf` — followed your design, including the hook.

**New files**
- `useAuthAnalytics.js` — owns `flowIdRef`/`attemptIdsRef`/`flowOutcomeRef`, the four `fire*` functions, and the shared `ssoError` state.
- `ForgotPasswordScreen.js` — the whole `AUTH_MODE.FORGOT_PASSWORD` screen.
- `authErrorMessages.js`, `AuthTextInput.js` — the pure message mappers and the input, which were the other things making the file long.

`AuthPage.js`: **806 → 409 lines.**

**On the catch you flagged** — the hook is instantiated exactly once, in `AuthPage`, and its outputs are passed down as props. `ForgotPasswordScreen` never calls `useAuthAnalytics` itself, which is the thing that would have quietly minted a second `flow_id` and split the funnel. `handleSSOTokenReceived`/`handleSSOError` also stayed in `AuthPage` rather than moving, since they close over `useSSOSignIn`'s success path — passed down as props too.

Every previous export is still re-exported from `'./AuthPage'`, so no test imports changed and no assertions were touched. Full suite is unchanged against baseline: 238/256, with the same 6 pre-existing unrelated failures (`FilterSearcher`, `history`, `AccountNavigationMenu`, `SettingsPage`, `SwipeableCategoryList`, `ReaderApp`).

One thing worth flagging since it came out of my own review rather than the tests: several comments moved with their code but kept positional references — "see the mount effect above", "hasUnrenderedEmailError below" — that no longer pointed at anything after the split. Repointed them at the owning module. Tests could never have caught that, and stale comments are worse than none.

YishaiGlasner (approved):

PR #218, opened 2026-07-22
fix(markdown links): resolve relative hrefs before opening

YishaiGlasner:
react-native-markdown-display passes anchor hrefs through unresolved, unlike RenderHTML which normalizes them against about:blank. Relative links in SimpleMarkdown (e.g. AboutBox descriptions) were reaching openUri/Linking.openURL as bare paths and failing to open.


Misc.js
```diff
@@ -1263,6 +1263,10 @@ SimpleHTMLView.propTypes = {
 const SimpleMarkdown = ({children, lang, style, markdownProps}) => {
   const {handleOpenURL} = useContext(ReaderAppContext);
   const onLinkPress = (url) => {
+    if (!/^[a-z][a-z0-9+.-]*:/i.test(url)) {
+      // markdown-display gives us the raw href with no resolution, unlike RenderHTML
+      url = Sefaria.api._baseHost.replace(/\/$/, '') + (url.startsWith('/') ? '' : '/') + url;
+    }
```
yodem: ## Code Review & Recommendations

### Pros

* **Good Edge Case Handling for Strings:** The custom string formatting handles relative paths cleanly regardless of whether they start with `/`.
* **Standard-Compliant Regex:** The scheme check correctly matches true URI schemes rather than just searching for `http(s)`.

### Areas for Improvement

#### 1. Performance (`useCallback`)

`onLinkPress` is instantiated on every render cycle. If passed as a prop to a complex Markdown component tree, this will cause unnecessary re-renders. Wrapping it in `useCallback` prevents this.

#### 2. Potential Crash on Undefined/Null URLs

If a markdown link is improperly formatted in the source text (e.g., `[]()`), `url` might be `undefined` or `null`, causing `regex.test()` or `url.startsWith()` to throw a runtime `TypeError`.

#### 3. Native `URL` API

Instead of manual string splitting and slash replacement, modern JavaScript environments (including modern React Native) support standard `URL` resolution:

```javascript
new URL(relativeUrl, baseHost).href

```

#### 4. Encapsulated API Property Access

Accessing `_baseHost` with a leading underscore (`_`) signals a private variable. Ideally, `Sefaria.api` should expose a public getter (e.g., `Sefaria.api.getBaseHost()`) to prevent tight coupling to internal implementation details.

---

## Refactored Code Example

Here is how you can write a safer, more performant, and complete version of the component:

```javascript
import React, { useContext, useCallback } from 'react';
import Markdown from 'react-native-markdown-display'; // Example target component
import { ReaderAppContext } from './ReaderAppContext';
import Sefaria from './Sefaria';

const SimpleMarkdown = ({ children, lang, style, markdownProps }) => {
  const { handleOpenURL } = useContext(ReaderAppContext);

  const onLinkPress = useCallback(
    (url) => {
      if (!url) return false;

      let resolvedUrl = url;
      const hasProtocol = /^[a-z][a-z0-9+.-]*:/i.test(url);

      if (!hasProtocol) {
        // Safe standard URL resolution
        const baseHost = Sefaria.api._baseHost; // Ideally Sefaria.api.getBaseHost()
        resolvedUrl = new URL(url, baseHost).href;
      }

      handleOpenURL(resolvedUrl);
      return false;
    },
    [handleOpenURL]
  );
```

yodem: ### Edge Cases & Pitfalls to Watch For

While this block gets the job done most of the time, here are four subtle ways it can break in production:

#### 1. Runtime Crash on Missing/Empty URLs

If a user writes a bad markdown link like `[Text]()` or `[Text]()` where `url` is `undefined`, `null`, or non-string:

```javascript
// Throws: TypeError: Cannot read property 'test' of undefined
!/^[a-z].../.test(url) 

```

> **Fix:** Check `if (!url)` early.

#### 2. Protocol-Relative URLs (`//example.com`)

If the markdown contains `[Link](//sefaria.org/Genesis)`, the regex `/^[a-z]...:`/ returns `false` because there is no letter before the slashes.

* Result: It appends `baseHost`, creating: `https://www.sefaria.org///sefaria.org/Genesis` (broken).

#### 3. Anchor Links (`#section-1`)

If the markdown contains internal section jumps like `[Jump](#notes)`:

* The regex returns `false`.
* Result: It constructs `[https://www.sefaria.org/#notes](https://www.sefaria.org/#notes)`, which will attempt a full page reload or open a new view instead of smooth-scrolling locally.

#### 4. Private Variable Access (`_baseHost`)

Accessing `_baseHost` (prefixed with `_`) risks breaking if the `Sefaria.api` module updates or refactors its private state.

---

### Recommended Replacement

If you want a modern, bulletproof rewrite using the built-in `URL` standard (available in modern JS & React Native engines like Hermes):

```javascript
// 1. Guard against empty/invalid types
if (!url || typeof url !== 'string') return false;

// 2. Allow hash-only anchor links to pass through unharmed
if (url.startsWith('#')) {
  handleOpenURL(url);
  return false;
}

// 3. Let URL constructor handle scheme resolution automatically
try {
  // If `url` is absolute (e.g., 'https://...', 'mailto:...'), the base parameter is ignored.
  // If `url` is relative (e.g., '/Genesis.1.1' or 'Genesis.1.1'), it resolves against Sefaria.api._baseHost.
  const resolved = new URL(url, Sefaria.api._baseHost).href;
  handleOpenURL(resolved);
} catch (e) {
  // Fallback for malformed URLs
  handleOpenURL(url);
}

```

yodem (dismissed):

yodem (approved):

PR #222, opened 2026-07-30
ci: provision SSO env vars for release builds

yodem:
## Problem

The app reads `GOOGLE_SSO_CLIENT_ID` and `GOOGLE_SSO_IOS_CLIENT_ID` through `react-native-config`, which bakes `.env` into the binary **at build time**. No workflow has ever created a `.env`, so CI-built apps ship with both values `undefined` — meaning Google Sign-In cannot work in a released build, independently of any OAuth-client configuration.

Verified by inspection: no `.env` is written anywhere under `.github/workflows/`, `fastlane/`, or `scripts/`.

## Change

Adds a `Create .env for react-native-config` step to the four deploy workflows (Android QA/release, iOS QA/release), immediately after `npm install` so it precedes every build.

## Why variables and not secrets

These are **public OAuth client identifiers**. They are extractable from any shipped APK/IPA by design — masking them in CI logs buys nothing, and treating them as secrets makes them harder for developers to obtain legitimately. They are therefore read from GitHub **repository variables**.

## Setup required before merge

Two repository variables must be configured (Settings → Secrets and variables → Actions → **Variables**):

| Variable | Value |
|---|---|
| `GOOGLE_SSO_CLIENT_ID` | the Google **Web** OAuth client ID |
| `GOOGLE_SSO_IOS_CLIENT_ID` | the Google **iOS** OAuth client ID |

Until they are set the step writes empty values, which is no worse than today's behaviour (absent).

## Notes

- `BASE_HOST` is deliberately **not** set, so `api.js` falls back to `https://www.sefaria.org/`. A release can therefore never be built pointing at a cauldron.
- No `APPLE_SSO_*` variable is included: none of them is referenced by any JS or native config in this repo — they are Django settings. In particular `APPLE_SSO_PRIVATE_KEY` must never be added here, since `react-native-config` would bake it into the shipped binary.
- Values are passed via an `env:` block and `printf` rather than interpolated directly into the `run:` script, avoiding shell injection on the expression.

https://claude.ai/code/session_01F2KcNAoRp7cdaf67qMdXDZ

### DIFF
--- a/.github/workflows/deploy-android-release.yml
+++ b/.github/workflows/deploy-android-release.yml
@@ -87,6 +87,19 @@ jobs:
       - name: Install npm dependencies
         run: npm install
 
+      # react-native-config bakes .env into the built binary at build time.
+      # These are public OAuth client identifiers (they are extractable from any
+      # shipped APK/IPA by design), so they live in GitHub *variables*, not
+      # secrets. BASE_HOST is deliberately NOT set here: api.js falls back to
+      # https://www.sefaria.org/ so a release can never point at a cauldron.
+      - name: Create .env for react-native-config
+        env:
+          GOOGLE_SSO_CLIENT_ID: ${{ vars.GOOGLE_SSO_CLIENT_ID }}
+          GOOGLE_SSO_IOS_CLIENT_ID: ${{ vars.GOOGLE_SSO_IOS_CLIENT_ID }}
+        run: |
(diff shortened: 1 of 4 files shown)


yodem:
Superseded by the work on `feature/sc-45083-sso-google-apple-signin`, which already writes `.env` in all four deploy workflows (Android QA/release, iOS QA/release) — sourcing the values from repository **secrets** rather than variables, and additionally writing `GOOGLE_SSO_IOS_CLIENT_ID` for the iOS builds. Closing as redundant; the branch is left in place.

PR #223, opened 2026-08-05
chore: first pass

stevekaplan123:
## Description
Moves every piece of interface text out of LocalizedStrings.js and into two JSON files, i18n/en.json and i18n/he.json, so that translators can edit Hebrew in Weblate instead of asking a developer to change code.

Before, all 235 English strings and their Hebrew translations lived in one 486-line JavaScript file. Weblate cannot read JavaScript. After, LocalizedStrings.js is 23 lines that just load the two JSON files, which Weblate can read and write directly. 

## The new way strings are named

Snake case is now used to bring continuity with the other Weblate implementations:
strings.aboutSefaria       →  strings.about.about_sefaria
strings.deleteAccountMsg   →  strings.account.delete_account_message
221 strings across 16 namespaces.

## Tests
i18n.test.js (10) — the dictionary check. Never renders. Key parity, snake_case ids, matching placeholders, no empty English, no identical en/he pair.
i18nRendering.test.js (80) — the screen check. Renders each screen twice, English and Hebrew, and reads back what a user would see.
Why both? account.account and nav.account are both "Account" in English but differ in Hebrew (חשבון משתמש vs חשבון). Use the wrong one and every dictionary test still passes. Only a Hebrew render catches it.

182 of 221 strings (82%) covered. The other 39 are each allowlisted with a written reason, and every one is a string no user can currently reach.


dcschreiber:
_(Claude writing on Daniel's behalf)_

Nice work — I checked the migration mechanically and it holds up: 236 keys, en/he sets identical, no placeholder mismatches, and every `strings.*` reference in the codebase resolves. Three comments:

**1. The identical-en/he test conflicts with the async translation workflow.** `i18n.test.js` asserts that no id has the same text in both languages. The moment a developer adds an English string before a translator has gotten to it in Weblate, CI goes red — which is exactly the asynchronous case the whole setup exists to support. Worth relaxing this or making it a warning.

**2. There's a behavior change bundled in here.** `SaveLine` gained a `heDref` prop and `SaveButton` now persists `he_ref` on saved items (the "fix topic bookmark undefined in string" commit). It looks like a genuine fix, but flagging it in case it slipped in unnoticed — it changes data written to users' reading history, which is well outside a strings refactor. Mentioning it in the description would help reviewers, though that's up to you.

**3. Commit/PR title.** "chore: first pass" is what ends up in history permanently after a squash-merge of a ~3,000-line refactor — worth renaming to something descriptive.

dcschreiber (commented):
Overall, this looks very good. Very hard to PR for obvious reasons, but I think it's ready basically to be approved. 
The one important note is the first comment Claude has: you can't have an English string without a Hebrew string. I just want to make sure that that's on purpose.

stevekaplan123:
> _(Claude writing on Daniel's behalf)_
> 
> Nice work — I checked the migration mechanically and it holds up: 236 keys, en/he sets identical, no placeholder mismatches, and every `strings.*` reference in the codebase resolves. Three comments:
> 
> **1. The identical-en/he test conflicts with the async translation workflow.** `i18n.test.js` asserts that no id has the same text in both languages. The moment a developer adds an English string before a translator has gotten to it in Weblate, CI goes red — which is exactly the asynchronous case the whole setup exists to support. Worth relaxing this or making it a warning.
> 
> **2. There's a behavior change bundled in here.** `SaveLine` gained a `heDref` prop and `SaveButton` now persists `he_ref` on saved items (the "fix topic bookmark undefined in string" commit). It looks like a genuine fix, but flagging it in case it slipped in unnoticed — it changes data written to users' reading history, which is well outside a strings refactor. Mentioning it in the description would help reviewers, though that's up to you.
> 
> **3. Commit/PR title.** "chore: first pass" is what ends up in history permanently after a squash-merge of a ~3,000-line refactor — worth renaming to something descriptive.

Good catch.  I've loosened the requirements since it is normal and acceptable to have an extra en.json key that isn't found in he.json.  I did make that behavior change since I found in QAing it that it was showing up as "undefined".

__tests__/i18n.test.js
```diff
@@ -0,0 +1,269 @@
...
+// Exactly two segments, not "at least two": `unflatten` in LocalizedStrings.js splits each id
+// on its first dot to build the object tree the string library needs, so a third segment
+// would land inside a leaf name and `strings.a.b.c` would stop resolving.
+const ID_RE = /^[a-z0-9_]+\.[a-z0-9_]+$/;
+
+// The files are flat -- {"common.ok": "OK"} -- so an id IS a key and no flattening is needed.
+// Namespaces are the distinct first segments.
+const NAMESPACES = [...new Set(Object.keys(en).map(id => id.split('.')[0]))];
+
+describe('i18n string files', () => {
```
dcschreiber, line 22: _(Claude writing on Daniel's behalf)_

**Question: how do these tests run in CI?**

This branch added `.github/workflows/i18n-tests.yml` and then removed it again in `9112bc4e`. Sefaria-Mobile has no other workflow that runs jest, and `gh pr checks 223` reports no checks at all on this branch. So these ~2,000 lines of tests only run when someone remembers to run them locally.

That matters most for the case the deleted workflow called out in its own header comment: Weblate commits translator edits straight into the repo and validates nothing, and a missing string renders blank rather than crashing. Those merge PRs are exactly where no human is reading the diff.

Was the removal deliberate, and is jest CI arriving another way? If not, could the workflow come back before this merges?

__tests__/i18n.test.js
```diff
@@ -0,0 +1,269 @@
...
+  test('every id is lowercase snake_case with exactly two segments', () => {
+    const bad = [...Object.keys(en), ...Object.keys(he)].filter(id => !ID_RE.test(id));
+    expect(bad).toEqual([]);
+  });
+
+  test('en.json and he.json define exactly the same ids', () => {
+    const enIds = Object.keys(en).sort();
+    const heIds = Object.keys(he).sort();
+    expect(heIds.filter(id => !(id in en))).toEqual([]);  // extra in he
+    expect(enIds.filter(id => !(id in he))).toEqual([]);  // missing from he
```
dcschreiber, line 32: _(Claude writing on Daniel's behalf)_

**Question: is strict key parity what you intended here?**

You wrote that an extra `en.json` key with no `he.json` counterpart is normal and acceptable. This test still requires the two files to match exactly in both directions, so that case turns CI red.

The runtime is already relaxed about it: `withFallback` walks the English keys and falls back for anything absent, so a missing Hebrew key renders English and nothing breaks. The test is stricter than the code it guards.

Small related point: if a key ever is missing from `he.json`, the placeholder test (line 51) and the coverage report (line 68) both call `he[id].trim()` and throw `TypeError: Cannot read properties of undefined` before this test can report the real problem. Guarding with `(he[id] || '')` keeps the failure readable.

dcschreiber (changes requested):
_(Claude writing on Daniel's behalf — these review comments were AI-generated and Daniel reviewed them before posting.)_

Re-reviewed after your last push. All three earlier points are addressed, and I checked the branch rather than taking it on trust:

- `withFallback` is the right answer to the untranslated-string problem. A missing or empty Hebrew value now renders the English source instead of a blank label.
- The migration is complete. No old-style `strings.camelCase` references are left in the source.
- I ran the suites locally: `i18n.test.js` + `i18nRendering.test.js` pass, 100 tests. The full suite has 14 failures across 5 suites; master has 18 across 6. So this branch adds no failures and fixes `AccountNavigationMenu`.

Two questions before approving, both inline.

__tests__/i18n.test.js
dcschreiber: Feel free to ignore this. Fixing this is out of scope.

__tests__/i18n.test.js
stevekaplan123: On parity: I checked Weblate directly. I added a test string in English, and it
wrote that key into he.json with an **empty value** rather than omitting it. So
the two files stay in parity on every sync, and strict parity is both intended
and free — nothing Weblate does on its own will trip it.

That also answers the "stricter than the code it guards" point. It is stricter,
deliberately. I deleted `common.ok` from `he.json` to check: `no id renders
empty in either language` still passes, because `withFallback` handles it — the
runtime genuinely doesn't care. What strict parity buys is detecting a *lost*
translation. If a bad merge or a hand edit drops a Hebrew string, subset rules
would let it silently fall back to English and nobody would notice. Parity turns
that into a failed build.

__tests__/i18n.test.js
stevekaplan123: Yes I removed it because Akiva said we are planning to work on github test integration for mobile in a future project.

dcschreiber (approved):

